PR #3, opened 2024-02-06
Setting up staging deploy worklfow

manasaV3:
## Description

Adds workflow to deploy to staging on merge to main for api

Depends on: https://github.com/chanzuckerberg/cryoet-infra/pull/85

### DIFF
--- a/.github/workflows/staging-deploy.yaml
+++ b/.github/workflows/staging-deploy.yaml
@@ -20,7 +20,7 @@ jobs:
       cancel-in-progress: true
 
     name: deploy staging branch
-    runs-on: ubuntu-latest
+    runs-on: [ARM64, self-hosted, Linux]
     environment: staging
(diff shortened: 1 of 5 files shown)


.github/workflows/staging-deploy.yaml
```diff
@@ -0,0 +1,48 @@
...
+      - ".happy/**"
+      - "api_server/**"
+
+# https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  deploy-to-staging:
```
jgadling, line 17: What do you think about trying to switch over ARM for the api? I already added a `platform: linux/arm64` line to the docker-compose.yml. So we either need to remove that, or add a `runs_on` parameter to this job to build on ARM, like so:

```
    runs-on: [ARM64, self-hosted, Linux]
```

jgadling (approved):

PR #5, opened 2024-02-06
[WIP] Markdown schema description for json metadata files. 

uermel:
This is a first pass at a markdown version of the data portal json metadata schema. It is a direct translation of the google doc. It does contain several errors, which I'll point out in comments below.

This should not be merged in current state because it is erroneous.


docs/v1.1.0/schema.md
```diff
@@ -0,0 +1,216 @@
...
+|                     |                      | Affiliation identifier      | affiliation_identifier       | STRING                                  |  --  |   --    |                    RECOMMENDED                    | ROR                                   | A unique identifier assigned to the affiliated institution by The Research Organization Registry (ROR)                                                                                                   |
+|                     |                      | Author list order           | order                        | INTEGER                                 |  --  |   --    |                       MUST                        | N/A                                   | The order in which the author appears in the publication                                                                                                                                                 |
```
uermel, line 40: The entire Author section is actually not a dictionary, but a list of Dictionaries of type "DatasetAuthor". We should reflect this in the documentation.

docs/v1.1.0/schema.md
```diff
@@ -0,0 +1,216 @@
...
+| Funding             | funding              | Funding agency name         | funding_agency_name          | STRING                                  |  --  |   --    |                    RECOMMENDED                    | ROR or Crossref Funder Registry       | Name of the funding agency. There could be multiple funding agencies                                                                                                                                     |
+|                     |                      | Grant ID                    | grant_id                     | STRING                                  |  --  |   --    |                    RECOMMENDED                    | N/A                                   | Grant identifier provided by the funding agency                                                                                                                                                          |
```
uermel, line 42: The entire Funding section is actually not a dictionary, but a list of Dictionaries of type "DatasetFunding". We should reflect this in the documentation.

docs/v1.1.0/schema.md
```diff
@@ -0,0 +1,216 @@
...
+|                     |                      | Related publication DOIs    | dataset_publications         | STRING<sup>**[1](#dataset_list)**</sup> |  --  |   --    |                     OPTIONAL                      | DOI                                   | DOIs of publications related to this dataset.                                                                                                                                                            |
+|                     |                      | Related database links      | related_database_links       | STRING<sup>**[1](#dataset_list)**</sup> |  --  |   --    |                     OPTIONAL                      | DOI                                   | Links to related database entries.                                                                                                                                                                       |
```
uermel, line 48: This actually turned out to be difficult in case of the strain ID, as each host would need to be validated. I think it would be better to create a list of allowable database prefixes in the related_database_entries field and rules for how to build URLs to them.

docs/v1.1.0/schema.md
```diff
@@ -0,0 +1,216 @@
...
+| Organism            | organism           | Organism name            | name           | STRING[^2]      | N/A    | N/A      | RECOMMENDED |                                                                                                                                | Name of the organism from which a biological sample used in a CryoET study is derived from, e.g. homo sapiens (reference)                                                                           |
+|                     |                    | Taxonomy ID              | taxonomy_id    | STRING          | N/A    | N/A      | RECOMMENDED | NCBI taxonomy ID                                                                                                               | NCBI taxonomy identifier for the organism, e.g. 9606 (reference)                                                                                                                                    |
```
uermel, line 64: Would be good to actually be consistent and use the [NCBITaxon](https://obofoundry.org/ontology/ncbitaxon.html) ontology instead of straight IDs, to be consistent across the database.

docs/v1.1.0/schema.md
```diff
@@ -0,0 +1,216 @@
...
+|---------------------|--------------------|--------------------------|----------------|-----------------|--------|----------|-------------|--------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Sample type         | sample_type        | N/A                      | N/A            | ENUMERATION[^1] | N/A    | N/A      | MUST        | Cell, tissue, organism, Intact organelle, In-vitro mixture of macromolecules or their complex, In-silico synthetic data, Other | Type of samples used in a CryoET study.                                                                                                                                                             |
```
uermel, line 62: This ENUM should be clearly defined separately.

docs/v1.1.0/schema.md
```diff
@@ -0,0 +1,216 @@
...
+| Visualization default         | is_visualization_default | N/A                               |                             | BOOLEAN    |  --  |  false  | Not by data submitter                                                 | Default: false                                                    | Annotation shown in visualization applications by default (e.g. neuroglancer)                                                                                                                                                                                                                                                      |
+
+## 6. Annotation File Metadata
+
+### 6.1 Point
+Point files are newline-delimited json files ([ndjson](https://clue.engineering/2018/introducing-reactphp-ndjson)). Each
+line contains a nested dictionary. 
+
+| Data elements | metadata key 1 | Sub-elements | metadata key 2 | Value type | Unit | Default | Requirement | Controlled vocabularies | Description                                                                      |
```
uermel, line 194: Is lacking the 'type' field that is actually present in the files.

docs/v1.1.0/schema.md
```diff
@@ -0,0 +1,216 @@
...
+|               |                | y            | y              |  FLOAT     | pix  |   --    | MUST        |                         | x coordinate of the point in pixels, origin is the pixel origin of the 3D image. |
+|               |                | z            | z              |  FLOAT     | pix  |   --    | MUST        |                         | x coordinate of the point in pixels, origin is the pixel origin of the 3D image. |
+
+### 6.2 OrientedPoint
+OrientedPoint files are newline-delimited json files ([ndjson](https://clue.engineering/2018/introducing-reactphp-ndjson)). Each
+line contains a nested dictionary. 
+
+| Data elements | metadata key 1       | Sub-elements | metadata key 2 | Value type            | Unit | Default | Requirement   | Controlled vocabularies | Description                                                                      | 
+|---------------|----------------------|--------------|----------------|-----------------------|:----:|:-------:|---------------|-------------------------|----------------------------------------------------------------------------------|
```
uermel, line 205: Is lacking the 'type' field that is actually present in the files.

docs/v1.1.0/schema.md
```diff
@@ -0,0 +1,216 @@
...
+| Annotation object state       | ???                      | N/A                               |                             | STRING     |  --  |   --    | OPTIONAL                                                              |                                                                   | Molecule state annotated (e.g. open, closed)                                                                                                                                                                                                                                                                                       |
```
uermel, line 176: In the google schema these are top level keys, but in the files they are nested.

docs/v1.1.0/schema.md
```diff
@@ -0,0 +1,216 @@
...
+| Tomogram file name            | ???                          | N/A                         |                             | STRING                     |  --   |   --    | MUST                                                                  | =Tilt_experiment.name+Processing+Processing_software+\[no\]Ctf | Tomogram file name                                                                                                                                                                                                 | 
+| Tilt series file name         | ???                          | N/A                         |                             | reference                  |  --   |   --    | MUST                                                                  |                                                                | The name of the tilt series used to construct the tomograms                                                                                                                                                        |
```
uermel, line 121: These exist in the google schema, but not in the implementation.

docs/v1.1.0/schema.md
```diff
@@ -0,0 +1,216 @@
...
+|                               |                              | Affiliation identifier      | affiliation_identifier      | STRING                     |  --   |   --    | RECOMMENDED                                                           | ROR                                                            | A unique identifier assigned to the affiliated institution by The Research Organization Registry (ROR).                                                                                                            |
+|                               |                              | Author list order           | author_list_order           | INTEGER                    |  --   |   --    | MUST                                                                  |                                                                | The order in which the author appears in the publication                                                                                                                                                           |
```
uermel, line 146: The entire Author section is actually not a dictionary, but a list of Dictionaries of type "TomogramAuthor". We should reflect this in the documentation.

docs/v1.1.0/schema.md
```diff
@@ -0,0 +1,216 @@
...
+|                               |                          | Author list order                 | author_list_order           | INTEGER    |  --  |   --    | MUST                                                                  |                                                                   | The order in which the author appears in the publication                                                                                                                                                                                                                                                                           |
```
uermel, line 164: The entire Author section is actually not a dictionary, but a list of Dictionaries of type "AnnotationAuthor". We should reflect this in the documentation.

docs/v1.1.0/schema.md
```diff
@@ -0,0 +1,216 @@
...
+| Authors                       | authors                  | ORCID                             | ORCID                       | STRING     |  --  |   --    | RECOMMENDED                                                           | ORCID                                                             | A unique, persistent identifier for the annotator, provided by ORCID                                                                                                                                                                                                                                                               |
```
uermel, line 156: 'ORCID' is the only metadata key that is capitalized. Should probably be lower case in the future.

docs/v1.1.0/schema.md
uermel, line 40: The 'Author list order' is not actually a field in the JSON files, it is explictely given by the order in the list (JSON arrays are ordered).

docs/v1.1.0/schema.md
uermel, line 146: The 'Author list order' is not actually a field in the JSON files, it is explictely given by the order in the list (JSON arrays are ordered).

docs/v1.1.0/schema.md
uermel, line 164: The 'Author list order' is not actually a field in the JSON files, it is explictely given by the order in the list (JSON arrays are ordered).

docs/v1.1.0/schema.md
jgadling: I don't know if we want to accept an explicit list order vs making that implied by the order that the authors are listed in the metadata file. Personally I'd prefer to make the ordering in the file significant.

jgadling: Yeah, I think we shouldn't include the field in the metadata spec and just say that the ordering is significant.

docs/v1.1.0/schema.md
```diff
@@ -0,0 +1,216 @@
...
+|-------------------------------|------------------------------|-----------------------------|-----------------------------|----------------------------|:-----:|:-------:|-----------------------------------------------------------------------|----------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Tomogram file name            | ???                          | N/A                         |                             | STRING                     |  --   |   --    | MUST                                                                  | =Tilt_experiment.name+Processing+Processing_software+\[no\]Ctf | Tomogram file name                                                                                                                                                                                                 |
```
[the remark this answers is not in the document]
jgadling: Should we drop this field? it seems moot?

docs/v1.1.0/schema.md
jgadling: This field seems useful - seems we should add it to our implementation?

docs/v1.1.0/schema.md
jgadling: Can we add that to this spec?

jgadling (approved):
I'd like to see the `author list order` and `tiltseries.tomogram` fields dropped if that's ok with you.

PR #15, opened 2024-02-23
Fix ingestion of segmentation masks

uermel:
Fixing https://github.com/chanzuckerberg/cryoet-data-portal/issues/502

1. The labels in segmentation masks were not correctly separated because `pyramid_to_mrc` in `common/image.py` took the original data, not the one from the pyramid (only the pyramid data is processed)

2. AnnotationSource classes in `importers/annotation.py` expected the keyword argument `is_viz_default`, but the YAML dict has `is_visualization_default`

3.  Import of `Point` annotations failed due to faulty indenting in `Point.load` in `importers/annotation.py`

4. Voxel spacing of the ingested segmentation masks was not correct, because `scale_maskfile` in `common/image.py` was not yet using the new override mechanism for the `voxel_spacing` (the correct `voxel_spacing` was actually passed inside `AnnotationImporter.import_annotations`, but not taken into account in the `SemanticSegmentationMaskFile.convert` method.

### DIFF
--- a/ingestion_tools/scripts/common/image.py
+++ b/ingestion_tools/scripts/common/image.py
@@ -217,6 +217,7 @@ def scale_maskfile(
     label: int,
     scale_z_axis: bool = True,
     write: bool = True,
+    voxel_spacing=None,
 ):
     mc = MaskConverter(fs, tomo_filename, label)
     pyramid = mc.make_pyramid(scale_z_axis=scale_z_axis)
@@ -226,6 +227,7 @@ def scale_maskfile(
         pyramid,
         f"{output_prefix}.mrc",
         write,
+        voxel_spacing=voxel_spacing,
     )
 
 
--- a/ingestion_tools/scripts/common/image.py
+++ b/ingestion_tools/scripts/common/image.py
@@ -71,7 +71,7 @@ class TomoConverter:
         mrcfiles.append(os.path.basename(filename))
 
         if write:
-            newfile = mrcfile.new(filename, self.data, overwrite=True)
+            newfile = mrcfile.new(filename, pyramid[0], overwrite=True)
             self.update_headers(newfile, header_mapper, voxel_spacing)
             newfile.flush()
             newfile.close()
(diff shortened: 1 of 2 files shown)


ingestion_tools/scripts/common/image.py
```diff
@@ -71,7 +71,7 @@ def pyramid_to_mrc(
         mrcfiles.append(os.path.basename(filename))
 
         if write:
-            newfile = mrcfile.new(filename, self.data, overwrite=True)
+            newfile = mrcfile.new(filename, pyramid[0], overwrite=True)
```
uermel, line 74: Using the processed mask data instead of original data.

ingestion_tools/scripts/importers/annotation.py
```diff
@@ -170,16 +170,16 @@ def load(
...
+            if skip_header:
+                next(points)
+            for coord in points:
+                annotation_set.append(
+                    self.point(
+                        float(coord[coord_order[0]]),
+                        float(coord[coord_order[1]]),
+                        float(coord[coord_order[2]]),
+                    ),
+                )
```
uermel, line 183: Fix indentation so the file is actually open.

ingestion_tools/scripts/importers/annotation.py
```diff
@@ -120,7 +120,8 @@ def __init__(
 
     def convert(self, fs: FileSystemApi, input_prefix: str, output_prefix: str, voxel_spacing: float = None):
         input_file = self.get_source_file(fs, input_prefix)
-        return scale_maskfile(fs, self.get_output_filename(output_prefix), input_file, self.label, write=True)
+        return scale_maskfile(fs, self.get_output_filename(output_prefix), input_file, self.label, write=True,
+                              voxel_spacing=voxel_spacing)
```
uermel, line 124: Add voxel_spacing override for segmentation masks.

ingestion_tools/scripts/importers/annotation.py
```diff
@@ -85,12 +85,12 @@ def __init__(
...
         glob_vars: dict[str, str],
         file_format: str,
-        is_viz_default: bool = False,
+        is_visualization_default: bool = False,
     ):
         self.glob_string = glob_string.format(**glob_vars)
         self.file_format = file_format
         self.shape = shape
-        self.is_viz_default = is_viz_default
+        self.is_viz_default = is_visualization_default
```
manasaV3, line 93: Can we also rename the class variable to `is_visualization_default` for consistency?

uermel: done with the last commit.

manasaV3 (approved):
:shipit:

PR #16, opened 2024-02-23
fix: Updating ingestor to use amd

manasaV3:
Making changes to revert the ingestor image to be built on `linux/amd` platform as there isn't sufficient support to use IMOD on the arm platform at the moment. 

Having the ingestor be a part of the docker-compose file used for building api-server causes build issues, so moving it to a separate docker config.

### DIFF
--- a/.github/workflows/ingestor-test.yml
+++ b/.github/workflows/ingestor-test.yml
@@ -13,7 +13,7 @@ on:
 jobs:
   py-test:
     name: "Run tests"
-    runs-on: ubuntu-latest
+    runs-on: [ARM64, self-hosted, Linux]
     steps:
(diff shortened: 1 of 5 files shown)


jgadling:
Does this need to be updated or does it actually work ok?
https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/main/ingestion_tools/Dockerfile#L13

manasaV3:
> Does this need to be updated or does it actually work ok?
> https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/main/ingestion_tools/Dockerfile#L13

The image built fine for amd with that config. I was able to validate in the image that imod commands were runnable.

jgadling (approved):

PR #20, opened 2024-03-01
Adding template for ingesting non-dataset submissions

manasaV3:
### Changes Introduced
- Adds template for ingesting data that is not a complete dataset

### DIFF
--- a/ingestion_tools/dataset_configs/submission_template.yaml
+++ b/ingestion_tools/dataset_configs/submission_template.yaml
@@ -1,0 +1,46 @@
+dataset:
+  dataset_identifier: REQUIRED #(ex: 10004)
+annotations:
+  - metadata:
+      annotation_object:
+        id: REQUIRED
+        name: REQUIRED
+        description: OPTIONAL
+        state: OPTIONAL
+      dates:
+        deposition_date: 2023-XX-XX
+        last_modified_date: 2023-XX-XX
+        release_date: 2023-XX-XX
+      annotation_method: REQUIRED
+      annotation_publications: OPTIONAL
+      ground_truth_status: REQUIRED #(true or false)
+      authors:
+        - name: RECOMMENDED
+          ORCID: 0000-XXXXXX
+          primary_annotator_status: OPTIONAL #(true or false)
+          corresponding_author_status: OPTIONAL #(true or false)
+      annotation_software: RECOMMENDED
+      version: 1.0
+      is_curator_recommended: OPTIONAL #(true or false)
+      confidence:
+        precision: REQUIRED IF ground_truth_used is specified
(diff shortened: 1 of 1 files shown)


ingestion_tools/dataset_configs/submission_template.yaml
```diff
@@ -0,0 +1,46 @@
+dataset:
+  dataset_identifier: REQUIRED #(ex: 10004)
+annotations:
```
manasaV3, line 3: Should the annotation be mapped to a tomogram or voxel spacing now?

ingestion_tools/dataset_configs/submission_template.yaml
```diff
@@ -0,0 +1,46 @@
...
+        glob_string: aaaa/bbb/{ts_name}.csv
+        shape: REQUIRED #(valid values: SemanticSegmentationMask, SegmentationMask, OrientedPoint, or Point)
+        is_visualization_default: OPTIONAL #(true or false)
+        delimiter: OPTIONAL
+        mask_label: OPTIONAL # (valid values: integer > 0, for SemanticSegmentationMask)
+runs: {}
+tiltseries: {} # To be expanded in future
+tomograms: {} # To be expanded in future
+standardization_config:
+  deposition_id: OPTIONAL
```
manasaV3, line 41: Can we use the depostion_id as a prefix to the imported annotation's naming?

anchi2c: I think yes, but we will need to add deposition_id also to the existing data so that they are consistent, I suppose.

manasaV3: We could default it to null for cases where no value is explicitly specified for it. :)

ingestion_tools/dataset_configs/submission_template.yaml
uermel: Annotations should be mapped to tomograms.

uermel (changes requested):
I advocate for adding the reference to source tomogram now, before we potentially have other tomos. Otherwise this looks good!

ingestion_tools/dataset_configs/submission_template.yaml
jgadling: They're currently mapped to voxel spacings -- should they be mapped directly to tomograms, or to voxel spacing + alignement?

ingestion_tools/dataset_configs/submission_template.yaml
jgadling: I think we may as well add a deposition identifier to existing datasets as well? They're still depositions, and handling nulls is often a mess.

jgadling:
Closing this in favor of the changes in progress in #34

PR #23, opened 2024-03-06
Add LinkML schema for metadata files and the ability to generate docs and pydantic classes from the schema.

uermel:
Common slots (shared between the database and metadata files) are defined in `schema/v1.1.0/common.yaml`, this includes type (called "range" in LinkML), description and optionally a unit or a regex pattern (e.g. for validation of identifiers).

The nested structure of the metadata files is defined in `schema/v1.1.0/metadata.yaml`. This file contains an "underdefined" version of the metadata schema, that contains references to `common.yaml` via the "exact_mappings" slot.

`schema.py` consumes `common.yaml` and `metadata.yaml` to produce `metadata_materialized.yaml`, which contains all all classes with correctly typed attributes with descriptions. This file can be used for code- and docgen. All generation is handled in the makefile. 

The README specifies the folder structure and links to the relevant file specs. We could change the doc templates if we wanted to. Recommend to start looking at this at the schema root directory: https://github.com/chanzuckerberg/cryoet-data-portal-backend/tree/uermel/linkml/schema/v1.1.0

### DIFF
--- a/schema/v1.1.0/Makefile
+++ b/schema/v1.1.0/Makefile
@@ -1,0 +1,13 @@
+.DEFAULT_GOAL := build
+
+.PHONY: clean
+clean:
+	rm -rf metadata-docs
+	rm -rf metadata_materialized.yaml
+	rm -rf metadata_models.py
+
+.PHONY: build
+build:
+	python schema.py materialize metadata.yaml common.yaml metadata_materialized.yaml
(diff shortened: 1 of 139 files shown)


uermel:
We can also think about generating the docs from the pydantic classes instead of the linkml materialized schema.

manasaV3 (approved):
Looks great! :fire:

anchi2c (approved):

anchi2c:
@manasaV3 I will leave this for you to merge if the checks it failed is fine.

PR #27, opened 2024-03-18
Add basic test for point annotations import

jgadling:
I'm hoping we can put everything back in a single docker-compose.yml file since it's way easier to work with. I'll fix the actions if they break.


jgadling:
Happy gets mad about `build` blocks in docker compose files with multiple platforms.

manasaV3 (approved):

PR #28, opened 2024-03-18
chore: Updating pre-commit config to run on modified files

manasaV3:
### Changes introduced: 
- Updates pre-commit to only run on files that have been updated.

### DIFF
--- a/.github/workflows/py-formatting.yml
+++ b/.github/workflows/py-formatting.yml
@@ -22,4 +22,4 @@ jobs:
       - name: check backend
         uses: pre-commit/action@v3.0.0
         with:
-          extra_args: --files ${{ steps.files_changed.outputs.files}}
+          extra_args: --files ${{ steps.files_changed.outputs.files }}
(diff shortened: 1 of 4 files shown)


uermel:
Instead of fixing the linting of the `metadata_models.py` directly it would probably be better to fix the makefile. I had attempted to include linting in the build process here: https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/757fd3f791969babf8bf53e9ba917cc62cfe2072/schema/v1.1.0/Makefile#L14 , but I think this only does black linting and doesn't pass ruff. I'm not sure what the correct ruff command would be to fix it appropriately?

manasaV3:
@uermel If the pre-commit is configured locally, it will run ruff, black, and the other hooks automatically before committing the changes. What are your thoughts on having the linting be a part of that instead of the build step?

uermel:
> @uermel If the pre-commit is configured locally, it will run ruff, black, and the other hooks automatically before committing the changes. What are your thoughts on having the linting be a part of that instead of the build step?

Fine with me.  Sorry for not considering this should be its own package.

manasaV3:
No worries! :)

jgadling (approved):

PR #29, opened 2024-03-19
Fix downscaling and data type of segmentation masks

uermel:
After the initial comparison with the label the maps are already binary, so after the scaling operation we need to compare with 1 instead of comparing with the label again. 

Scaled masks are binary and should be stored as int8.

This PR fixes that.

### DIFF
--- a/ingestion_tools/scripts/common/image.py
+++ b/ingestion_tools/scripts/common/image.py
@@ -148,7 +148,7 @@ class MaskConverter(TomoConverter):
 
             # For semantic segmentation masks we want to have a binary output.
             # downscale_local_mean will return float array even for bool input with non-binary values
-            scaled = (downscale_local_mean(pyramid[i] == 1, (z_scale, 2, 2)) > 0).astype(np.int8)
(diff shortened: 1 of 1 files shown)


ingestion_tools/scripts/common/image.py
```diff
@@ -148,7 +148,7 @@ def make_pyramid(self, max_layers: int = 2, scale_z_axis: bool = True) -> List[n
 
             # For semantic segmentation masks we want to have a binary output.
             # downscale_local_mean will return float array even for bool input with non-binary values
-            scaled = (downscale_local_mean(pyramid[i] == self.label, (z_scale, 2, 2)) > 0).astype(np.float32)
+            scaled = (downscale_local_mean(pyramid[i] == 1, (z_scale, 2, 2)) > 0).astype(np.int8)
```
manasaV3, line 151: Could this be just be pyramid?

uermel: You are right.

manasaV3 (approved):

PR #31, opened 2024-03-22
Add Annotation Importer for Instance segmentations. 

uermel:
Imports TARDIS instance points, new point formate with the additional label "instance_id".


uermel:
@jgadling

In creating this for the short term I have some thoughts on how we import point annotations:
- ideally the voxel_spacing should be available to the Importer.load method so point coordinates can be scaled by the voxel_size should they be given in Angstrom
- we should consolidate all point importers and maybe follow the pattern I suggest in this PR (a dataclass defining the point format)

manasaV3 (approved):

jgadling (approved):

PR #33, opened 2024-03-25
feat: Add graphql prod deploy workflow

manasaV3:
### Changes introduced: 

Adds GitHub workflow to deploy GraphQL to prod environment.

### DIFF
--- a/.github/workflows/prod-deploy.yaml
+++ b/.github/workflows/prod-deploy.yaml
@@ -9,9 +9,9 @@ permissions:
 
 jobs:
   deploy-to-prod:
-    concurrency:
-      group: prod-cryoet-api-deploy-${{ github.ref }}
-      cancel-in-progress: true
+#    concurrency:
+#      group: prod-cryoet-api-deploy-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/prod-deploy.yaml
```diff
@@ -0,0 +1,42 @@
...
+# https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  deploy-to-prod:
+    concurrency:
+      group: prod-cryoet-api-deploy-${{ github.ref }}
+      cancel-in-progress: true
```
jgadling, line 14: Curious why this is here?

jgadling (approved):

.github/workflows/prod-deploy.yaml
manasaV3: I retained it from the copy pasta, to handle the case where we accidentally trigger multiple executions of prod deployments manually.

PR #34, opened 2024-03-26
Support depositions of annotations that are associated with existing datasets

jgadling:
This is the very beginning of a refactor to change the way we "find" annotations, tomograms, etc in our standardization script. It has just-enough functionality to support the current set of annotations that we need to import, but we'll be continuing this code refactor (and ingestion config refactor!) right after the current import is complete.


uermel (approved):
👍

PR #44, opened 2024-04-10
chore: Adding method_type to config template

manasaV3:
Adding `annotation.method_type` to dataset config 10301.yaml


uermel (approved):
Thank you!

PR #45, opened 2024-04-10
Ingestion config for 10302 and required fixes

uermel:
- First stab at ingestion config for 10302
- changes to tomoman_to_tsv.py concerning this dataset: gainref location, check for presence of defect file entry

- fix for 10301: location of some rawtlt files.

### DIFF
--- a/ingestion_tools/dataset_configs/10302.yaml
+++ b/ingestion_tools/dataset_configs/10302.yaml
@@ -117,14 +117,14 @@ tomograms:
 
 standardization_config:
   destination_prefix: "10302"
-  source_prefix: "cryoetportal-rawdatasets-dev/chlamy2/data/"
+  source_prefix: "chlamy2/data/"
   run_data_map_file: "chlamy_visual_proteomics/run_to_data_map.tsv"
-  run_glob: "*"
-  run_regex: /\d{8}_.*$
+  run_glob: "chlamy_visual_proteomics/*"
+  run_regex: /(\d{8}_.*)$
   run_name_regex: (.*)
-  frames_glob: null # "chlamy_visual_proteomics/{run_name}/frames/*.eer"
-  frames_name_regex: null #(.*)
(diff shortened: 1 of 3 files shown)


ingestion_tools/dataset_configs/10302.yaml
```diff
@@ -117,14 +117,14 @@ tomograms:
...
-  run_regex: /\d{8}_.*$
+  run_glob: "chlamy_visual_proteomics/*"
+  run_regex: /(\d{8}_.*)$
   run_name_regex: (.*)
-  frames_glob: null # "chlamy_visual_proteomics/{run_name}/frames/*.eer"
-  frames_name_regex: null #(.*)
-  gain_glob: null # "chlamy_visual_proteomics/{frame_gain_reference}"
+  #frames_glob: null # "chlamy_visual_proteomics/{run_name}/frames/*.eer"
+  #frames_name_regex: null #(.*)
+  #gain_glob: null # "chlamy_visual_proteomics/{frame_gain_reference}"
```
uermel, line 127: @manasaV3 Currently, the test import fails due to the frames_glob missing. It also failed when I set it to `null`. What is the correct value here to skip frames?

uermel: Using an empty string worked.

ingestion_tools/dataset_configs/10302.yaml
```diff
@@ -0,0 +1,140 @@
+dataset:
+  authors: &dataset_authors
+    - name: Ron Kelley,
+      primary_author_status: true
+    - name: Sagar Khavnekar,
```
manasaV3, line 5: Can we remove the commas here as well?

manasaV3 (approved):

ingestion_tools/dataset_configs/10302.yaml
uermel: Done.

PR #48, opened 2024-04-11
feat: Updating enqueue to support skip-until

manasaV3:
### Changes introduced
Adds support to skip runs until a specific run name.


ingestion_tools/scripts/enqueue_runs.py
```diff
@@ -219,7 +227,15 @@ def queue(
     # Always iterate over datasets and runs.
     dataset = DatasetImporter(config, None)
     digitmatch = re.compile(r"[^\d]+(\d+)[^\d]+")
+
+    skip_run = skip_until_run_name is not None
+    skip_run_until_regex = re.compile(skip_until_run_name)
```
jgadling, line 232: This is going to explode if we don't specify a skip-until-run-name argument

jgadling (approved):

PR #49, opened 2024-04-11
chore: Making annotation object name consistent

manasaV3:
Relates to: https://github.com/chanzuckerberg/cryoet-data-portal/issues/396


uermel (approved):
👍

PR #51, opened 2024-04-12
chore: Make a new `annotation_authors.primary_author_status` field

jgadling:
This field is set up to have the exact same data as `annotation_authors.primary_annotation_status` so we can add the new field, get clients updated to use it, and then deprecate the old one.

### DIFF
--- a/api_server/migrations/cryoetdataportal/1712956764128_alter_table_public_annotation_authors_add_column_primary_author_status/down.sql
+++ b/api_server/migrations/cryoetdataportal/1712956764128_alter_table_public_annotation_authors_add_column_primary_author_status/down.sql
@@ -1,4 +1,1 @@
--- Could not auto-generate a down migration.
--- Please write an appropriate down migration for the SQL below:
--- alter table "public"."annotation_authors" add column "primary_author_status" boolean
---  null;
+alter table "public"."annotation_authors" drop column "primary_author_status";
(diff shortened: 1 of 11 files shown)


api_server/migrations/cryoetdataportal/1712956764128_alter_table_public_annotation_authors_add_column_primary_author_status/up.sql
```diff
@@ -0,0 +1,3 @@
+alter table "public"."annotation_authors" add column "primary_author_status" boolean
+ null;
+UPDATE "public"."annotation_authors" SET primary_author_status = primary_annotator_status;
```
manasaV3, line 3: 🔥

api_server/migrations/cryoetdataportal/1712956764128_alter_table_public_annotation_authors_add_column_primary_author_status/down.sql
```diff
@@ -0,0 +1,4 @@
+-- Could not auto-generate a down migration.
```
manasaV3, line 1: Can we update this query or delete this file?

manasaV3 (commented):
The spacing formatting looks like it is setting the line length to 80, from 120.  Is that intentional? :)

manasaV3 (approved):

PR #63, opened 2024-05-13
Refactor and correct point annotation import code. 

uermel:
Adresses https://github.com/chanzuckerberg/cryoet-data-portal/issues/703

This PR encompasses: 

- refactor point annotation classes (point, oriented point, instance segmentation) code to rely on same base class
- define dataclasses for point, orientedPoint and instancePoint instead of defining this in methods
- refactor reading/conversion functions to be in one location
- compute the location and orientation correctly 
- add support for importing mod files. 
- test cases and test files for import of all supported point file types.

### Details
**Refactor point annotation classes**
`importers.annotation.AbstractPointFile` defines all methods that can be shared between the point annotation classes. This way, `PointFile` and `OrientedPointFile` only have a custom init and don't reimplement functionality. `AbstractPointFile` has a class-attribute `map_functions` that maps `file_format` to the conversion function. `InstanceSegmentationFile` has a custom `init` and `get_object_count`.

**Define dataclasses for point, orientedPoint and instancePoint instead of defining this in methods**
These are defined in `common.point_coverter`. All share the `to_dict`-method that defines one line of the ndjson-output. 

**refactor reading/conversion functions to be in one location**
All readers now live in `common.point_coverter`. This allows reusing the same csv-converter for points and instance segmentations

**compute the location and orientation correctly**
Given a tomogram-array of dimensions `(dim_z, dim_y, dim_x)` we currently store the locations as `location['x']` for `dim_x`, `location['y']` for `dim_y` and  `location['z']` for `dim_z`. In the current version, the points are

Point:
-read as xyz directly

Oriented Point: 
- read into a numpy array as [N, 3] in order `[x, y, z]` in `from_relion3_star`, `from_relion4_star`, etc
- flipped to `[z, y, x]` in `invert_positions_axis_order`
- flipped back to `[x, y, z]` in OrientedPointFile.oriented_point

InstanceSegmentation: 
- read as xyz directly

This PR changes the config files for the STAR formats such that points are read directly as xyz instead of flipping twice.

The orientations should also be given in a standard rotation matrix (as output for scipy.transform.Rotation instead of flipping the output (as done currently).

**add support for importing mod files.** 
Adds the IMOD format as a Point-Annotation file.

**test cases and test files for import of all supported point file types**
added tests for ingestion of `csv`, `csv_with_header`, `relion3_star`, `relion4_star`, `tomoman_relion_star`, `stopgap_star`, `mod`, `tardis`.


ingestion_tools/scripts/common/point_converter.py
```diff
@@ -0,0 +1,368 @@
...
+                "x": self.x_coord,
+                "y": self.y_coord,
+                "z": self.z_coord,
+            },
+        }
+
+
+@dataclass
+class OrientedPoint:
+    x_coord: float
```
manasaV3, line 36: This and the InstancePoint could inherit from the `Point` class.

manasaV3 (approved):

PR #65, opened 2024-05-20
Refactor the way we handle dependencies in the importer

jgadling:
OK, this is a huge PR, and it does several things to try to standardize the way we handle each of the different classes of data that the portal supports

Dataset config changes:
- Each object type (tomograms, tiltseries, etc) now supports a **list** of configurations. 
- each item in that list is broken down into a block of `metadata`, and a list of `sources`
- the `sources` blocks contain a list of configurations that tells the ingestion script how to **find** the data that needs to be ingested.
- The standardization script will iterate over **all** items in the list of child configurations for each parent configuration (for example, when we find a `run`, **all** `tiltseries` configurations will be evaluated to see if we can find valid tiltseries for that run. 
- each source block allows for a `parent_filters` block to be defined to prevent that source block from looking for data when a parent object matches it's include/exclude criteria. so we can configure a given `tomogram` configuration to only be evaluated when its `run` parent is named `RUN001` for example
- What's a little less clear about supporting multiple object configurations, and parent-filtering, is that we can use this functionality to specify per-run/per-dataset/per-whatever overrides!

Standardization script changes:
- The standardization script is now aware of the data hierarchy that we need to import. It recurses over the hierarchy to find and ingest data
- The standardization script now generates most of the CLI arguments dynamically based on the data hierarchy

Finders:
- Each object type specifies a class that knows how to process the valid `source` type for that type. There is a standard set of `source` types supported by most objects - globbing for files, supplying literal values, etc
- some classes need fancier finders - like we need the ability to read voxel spacings from tomogram headers for example. we create custom finders for these types, but since all finders have the same interface, we don't need to do a lot of work / refactoring to support new/different finder methods.

Other notes:
- There's a `transform-ingestion-configs.py` script that can convert old-style dataset ingestion configs into new-style dataset ingestion configs.
- The `enqueue_runs.py script has been updated to work with some of the changes made to `standardize_dirs.py`, such as auto-generating CLI options.


ingestion_tools/scripts/importers/annotation.py
```diff
@@ -24,28 +35,199 @@ class AnnotationObject(TypedDict):
...
     sources: list[AnnotationSource]
 
 
-class BaseAnnotationSource:
+# This class is basically a global var that lets us cache metadata and identifiers for annotations,
+# so we can generate non-conflicting sequential identifiers for annotations as they're imported.
+class AnnotationIdentifierHelper:
+    next_identifier: dict[str, int] = defaultdict(partial(int, 100))
+    cached_identifiers: dict[str, int] = {}
+    loaded_vs_metadatas: dict[str, bool] = {}
```
manasaV3, line 57: Could this be a set?

ingestion_tools/scripts/importers/dataset_key_photo.py
```diff
@@ -16,12 +16,14 @@
 
 class DatasetKeyPhotoImporter(BaseImporter):
     type_key = "dataset_keyphoto"
+    plural_key = "dataset_keyphotos"
     image_keys = ["snapshot", "thumbnail"]
+    finder_factory = DefaultImporterFactory
```
manasaV3, line 21: Does this handle cases where no link or sources are specified for `dataset_keyphotos` in configs?

ingestion_tools/scripts/importers/voxel_spacing.py
```diff
@@ -0,0 +1,108 @@
...
+    ):
+        super().__init__(*args, **kwargs)
+        self.set_voxel_spacing(self.name)
+
+    # TODO mutating importers is bad news :'(
+    def set_voxel_spacing(self, voxel_spacing):
+        self.name = self.format_voxel_spacing(float(voxel_spacing))
+
+    @classmethod
+    def format_voxel_spacing(cls, voxel_spacing: float) -> None:
```
manasaV3, line 98: nit: return type

ingestion_tools/scripts/importers/voxel_spacing.py
```diff
@@ -0,0 +1,108 @@
...
+        if source.get("destination_glob"):
+            return DestinationGlobFinder(**source["destination_glob"])
+        if source.get("tomogram_header"):
+            return TomogramHeaderFinder(**source["tomogram_header"])
+        if source.get("literal"):
+            return VoxelSpacingLiteralValueFinder(**source["literal"])
+        raise Exception("Invalid source type")
+
+
+class VoxelSpacingImporter(BaseImporter):
```
manasaV3, line 79: This makes the whole processing of voxel spacing much more cleaner. 💜

ingestion_tools/scripts/importers/neuroglancer.py
```diff
@@ -78,5 +82,5 @@ def get_shader_controller(cls, tomo_header: np.recarray) -> dict[str, Any]:
         }
 
     @classmethod
-    def find_ng(cls, config: DataImportConfig, tomo: TomogramImporter) -> list["NeuroglancerImporter"]:
+    def find_ng(cls, config: DepositionImportConfig, tomo: TomogramImporter) -> list["NeuroglancerImporter"]:
         return [cls(config=config, parent=tomo)]
```
manasaV3, line 86: Is this finder still needed?

manasaV3 (approved):

ingestion_tools/scripts/importers/neuroglancer.py
jgadling: Yep, there's code in `config.py` that adds neuroglancer and key_images sections to all configs behind the scenes, so the key image and ng importers are definitely still needed!

ingestion_tools/scripts/importers/annotation.py
jgadling: Nope, it needs to be indexed by a specific set of annotation metadata, since it's basically a gross global variable that knows how to calculate identifiers for all annotations imported by a particular run of the import script :'( it's not a great design, but basically we need to share state between multiple different AnnotationImporter instances somehow.

ingestion_tools/scripts/common/fs.py
```diff
@@ -179,7 +179,7 @@ def destformat(self, path: str) -> str:
     def copy(self, src_path: str, dest_path: str) -> None:
         shutil.copy(src_path, dest_path)
 
-    def read_block(self, path: str, start: int, end: int) -> str:
+    def read_block(self, path: str, start: int = 0, end: int = 0) -> str:
```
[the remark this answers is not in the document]
jgadling: Actually we shouldn't be accepting scalar defaults here at all - we have several functions across this codebase that need fixing. :'(

ingestion_tools/scripts/importers/dataset_key_photo.py
jgadling: hehhh... actually it doesn't. Fixing!

PR #68, opened 2024-05-22
Fix docker push

jgadling:
the ECR-push github action was broken - this fixes it. I tested it on a branch a few minutes ago.

### DIFF
--- a/.github/workflows/push-ingestor-build.yaml
+++ b/.github/workflows/push-ingestor-build.yaml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-      - jesstest
     paths:
       - "ingestion_tools/**"
       - ".github/workflows/push-ingestor-build.yaml"
--- a/.github/workflows/push-ingestor-build.yaml
+++ b/.github/workflows/push-ingestor-build.yaml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - jesstest
     paths:
       - "ingestion_tools/**"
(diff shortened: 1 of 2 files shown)


.github/workflows/push-ingestor-build.yaml
```diff
@@ -27,10 +27,16 @@ jobs:
...
-        env:
-          ECR_REPO: ${{ secrets.ECR_REPO }}
-          AWS_REGION: ${{ secrets.AWS_REGION }}
-        run: make push-ingestor-build ecr_repo=$ECR_REPO tag=latest aws_region=$AWS_REGION
+      - name: Docker build & Push
+        uses: docker/build-push-action@v4
+        with:
+          context: ./ingestion_tools
+          file: ./ingestion_tools/Dockerfile
+          tags: ${{ secrets.ECR_REPO }}:${{ github.head_ref || github.ref_name }}
```
manasaV3, line 40: As this is triggered on push to main, `github.head_ref` might always turn up empty.

manasaV3 (approved):
Should we also set the tag in the enqueue script to default to `main`?

jgadling:
> Should we also set the tag in the enqueue script to default to `main`?

Good idea! I'll update it!

.github/workflows/push-ingestor-build.yaml
jgadling: I tested it on another branch, and the branch name came through, so 🤞 I'm expecting this to resolve to `main` in most cases

PR #77, opened 2024-05-30
Handle older MRC files that erroneously wrote uint8 data instead of int8

jgadling:
The simplest way I found to determine whether a tomogram has uint8 data that gets read by the mrcfile library as int8 is to check the `dmax` header value. Basically if the max value written to that header is >127, we've likely overflowed some data. It's possible that some older versions of the mrc writing library didn't write this header correctly, or wrote wrong values, but that seems unlikely.

What happens in the case of these older tomograms is that uint data (range 0:255) gets read in as int data (byte-order 0:127, and then -128:-1). This meant that the uint values (127, 128, 129) which were supposed to be contiguous, get mapped to (127,  -128, -127), which makes them discontiguous. This patch re-maps all of the incorrectly-read data back into a contiguous range from -128:127.

I still need to write some utilities to identify & fix the offending tomograms, but this is a start.

Some nice wikipedia reading: https://en.m.wikipedia.org/wiki/Signed_number_representations
The `mrcfile` library reads unsigned data into a two's-complement representation, and then this patch rearranges it so that it looks like it had been read as an offset-binary representation.


uermel:
The problem using the min/max values is that these are optional, so they might not be set or reflect the data contained. If this is not the case for any of our data we can use it of course, just something we should stay aware of as an edge case.

ingestion_tools/scripts/common/image.py
```diff
@@ -164,6 +164,13 @@ def get_mrc_extended_header(self) -> np.rec.array:
         return self.extended_header
 
     def get_pyramid_base_data(self) -> np.ndarray:
+        # We have some tomograms that were written before the 2014 spec was finalized and
+        # decided that MRC type 0 should store *signed* integers. The older mrc's store each
+        # voxel value as *unsigned* int8, while the MRC library reads that data as signed
+        # int8. Here we're remapping the overflowed/mangled uint8 data to a continuous
+        # range of int8 data
```
uermel, line 171: Can we note here that we're relying on the pre-computed header values?

uermel (approved):
Otherwise looks good.

manasaV3 (approved):

jgadling:
> The problem using the min/max values is that these are optional, so they might not be set or reflect the data contained. If this is not the case for any of our data we can use it of course, just something we should stay aware of as an edge case.

Yeah, unfortunately there's not a lot of other hints that we have to determine whether data has overflowed - the headers for "good" and "bad" tomograms look virtually identical otherwise, so we'd have to rely on processing the image data itself, or use the (also optional, and imnsho even less reliable) date information in the label header to make decisions.

PR #89, opened 2024-06-06
Fix cell type field names.

jgadling:
These field names were inaccurate and can't be read by the db import script


manasaV3 (approved):
Is there an update needed to the script as well?

PR #112, opened 2024-06-29
Run ingestion scripts semi-locally and update READMEs

daniel-ji:
Note that it is semi-locally since we still depend on the servers running in Docker containers.

To run ingestion scripts (for dev / debugging purposes) on your local machine, port 5433 is exposed to the local machine (arbitrary port) since trying to connect to port 5432 was giving issues, see https://stackoverflow.com/a/55259785. 

Also updated README files to provide a bit more insight to running tests in general, as well as adding sections about how to get Dockerized and local testing running (especially local testing, since there's some workarounds to get all tests passing with the postgresql / motoserver / nginx servers on Docker).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,21 @@
 # Community Tomography Data Portal Backend
 CryoET Portal API server and ingestion tools
 
-To launch a local dev environment:
+To launch a local dev environment (Docker required):
 ```
 make init
 ```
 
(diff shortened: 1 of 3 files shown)


daniel-ji:
Discussed with @jgadling, no need to have these local instructions, focus is on Docker support. IDE integration is already supported with VSCode as well. Thanks!

PR #116, opened 2024-07-05
Deposition config template update

manasaV3:
Relates to https://github.com/chanzuckerberg/cryoet-data-portal/issues/783

## Description
Based on the design interface, the required field additions to the metadata have been identified. The template.yaml has been updated to reflect the same.


uermel (approved):
Looks good to me.

ingestion_tools/dataset_configs/template.yaml
```diff
@@ -98,6 +102,17 @@ datasets: REQUIRED
...
     parent_filters: SEE runs.sources.parent_filters
+depositions:
+  - metadata: REQUIRED
+      authors: SEE annotations.authors
+      cross_references: SEE datasets.cross_references
+      dates: SEE annotations.dates
+      deposition_description: REQUIRED, STRING
+      deposition_identifier: REQUIRED, INTEGER
+      deposition_title: REQUIRED, STRING
+      type: REQUIRED, STRING (ENUM [TOMOGRAM, ANNOTATION, TOMOGRAM_ANNOTATION])
```
jgadling, line 113: This should probably be a list? a deposition can include tomograms, annotations, datasets, etc.

manasaV3: As the initial use case for this was was to distinguish the depositions that were just annotation, from depositions that contained other data as well, I had this be an enum for ease of use. 

What are your thoughts on this being an enum with values `[annotations, datasets, tomograms]`. I believe this encompasses all of use cases?

jgadling: I feel like making this an enum for just our current use case is just painting ourselves into a corner -- in reality, depositions can have any combination of data types! Unless the expectation is that this field will be deprecated once the depositions page supports more types of data?

jgadling (approved):

PR #118, opened 2024-07-06
Ingesting deposition metadata and keyphoto into s3 

manasaV3:
Relates to https://github.com/chanzuckerberg/cryoet-data-portal/issues/784

## Description
- Adds the deposition as the root for all ingestions. Handles metadata ingestion for the deposition, including the case where the metadata is empty. 
- Adds support for deposition key photo ingestion as well.


ingestion_tools/scripts/tests/s3_import/test_gains.py
```diff
@@ -8,7 +8,7 @@
 from importers.run import RunImporter
 from mypy_boto3_s3 import S3Client
 from standardize_dirs import IMPORTERS
-from tests.s3_import.test_annotations_import import list_dir
+from tests.s3_import.util import list_dir
```
jgadling, line 11: nit: looks like this is imported and not used?

jgadling (approved):

jgadling:
Does this mean we need to add deposition info to all configs ASAP?

manasaV3:
> Does this mean we need to add deposition info to all configs ASAP?

This [PR](https://github.com/chanzuckerberg/cryoet-data-portal-backend/pull/135) already adds the deposition entity to all the configs. :)

ingestion_tools/scripts/tests/s3_import/test_gains.py
manasaV3: It is used in the file, example [line](https://github.com/chanzuckerberg/cryoet-data-portal-backend/pull/118/files#diff-9dcb05a762b5ed051838d1a9cb70bc1b1acb4ada435d3c2d0c334647cc03d901R45).

PR #119, opened 2024-07-09
Cleaning up Jensen tiltseries without files

manasaV3:
## Description

Refactors entries for tiltseries in Jensen dataset. There are two cases that are handled here. 
1. None of the runs have a `tiltseries_path` defined. This means there are no valid tilt series for the dataset, so we remove the tilt series entry completely.
2. A few of the runs don't have `tiltseries_path` defined. We exclude those runs from the tiltseries processing.


ingestion_tools/dataset_configs/gjensen/run_data_map/10187.csv
```diff
@@ -50,5 +50,5 @@ cwl2011-08-13-23,IMOD,FIDUCIAL,Weighted back projection,IMOD,18.92,9.46,IMOD,4
 cwl2011-08-13-24,IMOD,FIDUCIAL,Weighted back projection,IMOD,18.92,9.46,IMOD,4
 cwl2011-08-13-25,IMOD,FIDUCIAL,Weighted back projection,IMOD,18.92,9.46,IMOD,4
 cwl2011-08-12-31,IMOD,FIDUCIAL,Weighted back projection,IMOD,12.521,6.26,IMOD,4
-cwl2011-08-13-8,IMOD,FIDUCIAL,Weighted back projection,IMOD,18.92,,IMOD,4
 cwl2011-08-13-12,,NON_FIDUCIAL,None,None,,9.46,,1
+cwl2011-08-13-8,IMOD,FIDUCIAL,Weighted back projection,IMOD,18.92,,,
```
manasaV3, line 54: TODO: Should remove cases like this from the run_data_map as well.

jgadling (approved):

PR #120, opened 2024-07-09
fix: Fix finders and add test coverage for them.

jgadling:
Fixes: https://github.com/chanzuckerberg/cryoet-data-portal/issues/842

## Description
This fixes the broken logic for include/exclude filters in ingestion configs. Basically it updates the logic so that if *any* exclude filter for *any* parent object is triggered, the object import will be canceled, and *at least one* of the include filters for each of the parent filters must match for the import to proceed.


daniel-ji (approved):
Looks good to me!

PR #124, opened 2024-07-10
Supporting different extensions for gains

manasaV3:
Relates to https://github.com/chanzuckerberg/cryoet-data-portal/issues/858

## Description
When gain file has a non `.dm4` extension, we copy the file over with the extension in the source file, and we continue to maintain the behaviour of retaining the `.mrc` extension for the `.dm4` files.

### DIFF
--- a/ingestion_tools/scripts/common/config.py
+++ b/ingestion_tools/scripts/common/config.py
@@ -247,7 +247,7 @@ class DepositionImportConfig:
             "tomogram": "{dataset_name}/{run_name}/Tomograms/VoxelSpacing{voxel_spacing_name}/CanonicalTomogram",
             "key_image": "{dataset_name}/{run_name}/Tomograms/VoxelSpacing{voxel_spacing_name}/KeyPhotos",
             "tiltseries": "{dataset_name}/{run_name}/TiltSeries",
-            "gain": "{dataset_name}/{run_name}/Frames/{run_name}_gain.mrc",
+            "gain": "{dataset_name}/{run_name}/Frames/{run_name}_gain",
             "frame": "{dataset_name}/{run_name}/Frames",
             "rawtilt": "{dataset_name}/{run_name}/TiltSeries",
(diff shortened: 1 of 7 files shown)


ingestion_tools/scripts/common/config.py
```diff
@@ -247,7 +247,7 @@ def resolve_output_path(self, key: str, obj: BaseImporter) -> str:
             "tomogram": "{dataset_name}/{run_name}/Tomograms/VoxelSpacing{voxel_spacing_name}/CanonicalTomogram",
             "key_image": "{dataset_name}/{run_name}/Tomograms/VoxelSpacing{voxel_spacing_name}/KeyPhotos",
             "tiltseries": "{dataset_name}/{run_name}/TiltSeries",
-            "gain": "{dataset_name}/{run_name}/Frames/{run_name}_gain.mrc",
+            "gain": "{dataset_name}/{run_name}/Frames/{run_name}_gain",
```
jgadling, line 250: should this be `.gain` instead of `_gain` ?

ingestion_tools/scripts/tests/s3_import/test_gains.py
```diff
@@ -0,0 +1,34 @@
...
+    runs = list(RunImporter.finder(config, dataset=datasets[0]))
+    gains = list(GainImporter.finder(config, dataset=datasets[0], run=runs[0]))
+    for gain in gains:
+        gain.import_item()
+
+    dataset_name = datasets[0].name
+    run_name = runs[0].name
+    prefix = f"output/{dataset_name}/{run_name}/Frames"
+    gain_files = [basename(item) for item in list_dir(s3_client, test_output_bucket, prefix)]
+    assert f"{run_name}_gain.gain" in gain_files
```
jgadling, line 34: Can we also assert that we have an `mrc` in gain_files? May as well test both cases?

jgadling (approved):
would be nice to add an extra check to the test, but lgtm!

PR #125, opened 2024-07-10
Updates to tomolist script

manasaV3:
## Description

Update the run map generation to refactor the gain reference. The dataset has the issue, that some of the gain references are incorrect as the files specified don't exist. Address this by mapping the gain references that have invalid paths, but same name of existing gain references, to the existing gain references.

### DIFF
--- a/ingestion_tools/scripts/tomoman_to_tsv.py
+++ b/ingestion_tools/scripts/tomoman_to_tsv.py
@@ -1,4 +1,5 @@
 import csv
+import os.path
 from dataclasses import asdict, dataclass
 
 import click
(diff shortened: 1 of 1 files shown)


jgadling:
I think some of the logic in here is a little redundant/confusing - can it be refactored, with some comments?

jgadling (changes requested):
I have questions!

ingestion_tools/scripts/tomoman_to_tsv.py
```diff
@@ -214,24 +225,71 @@ def cli(ctx):
...
+        return
+    # The gain_ref filenames are named with the date in them and are hence distinct. In some cases the file path
+    # might be incorrect but the filename might match a file existing in s3. So, we are also tracking the filename and
+    # the correct path for it.
+    gain_filename_ref = gain_ref.split("/")[-1]
+    if gain_filename_ref in in_s3:
+        return
+    full_path = os.path.join("cryoetportal-rawdatasets-dev/chlamy2/data/chlamy_visual_proteomics", gain_ref)
+    if fs.glob(full_path):
+        in_s3[gain_ref] = gain_ref
```
jgadling, line 243: I don't think we need to store this version in our cache at all -- just the mapping from the filename to the *correct* path is fine, and that also means that `get_gain_ref` only needs to check for the filename.

jgadling (approved):
I think this could be cleaned up a bit further, but the comments make it so I understand what we're trying to work around!

PR #126, opened 2024-07-10
Refactor annotation.sources to have class-specific children elements

daniel-ji:
- refactored config files with annotation.sources field to have class-specific children elements
- also did additional refactoring on the files by the refactoring:
	- primary_annotator_status -> primary_author_status
	- removed null values 
	- version (string) -> version (float)
- did similar (non-breaking / changing) refactors to the new 10426-10435 dataset configs
- created ingestion_tools/scripts/refactor_annotation_sources.py to assist with the refactoring
- created helper ingestion_tools/scripts/common/yaml_refactor.py to assist with file discovery (will be used later on in validation as well)
- updated tests with refactored changes as well 

Note that this should not affect the storage of the sources data / any part of the Postgres database schema aspect of the ingestion process, because the annotation importer has been modified to take the new dataset configuration schema structure and convert it to the same structure it looked like previously. (See ingestion_tools/scripts/importers/annotation.py)

Note: schema was not updated, will be updated in a later PR (see #128)

Addresses https://github.com/chanzuckerberg/cryoet-data-portal/issues/853


ingestion_tools/scripts/refactor_annotation_sources.py
```diff
@@ -0,0 +1,50 @@
...
+
+    for file in files_to_validate:
+        try:
+            with open(file, "r") as stream:
+                config_data = yaml.safe_load(stream)
+            file_changed = False
+            # If file has annotations, refactor sources
+            for entry in config_data["annotations"]:
+                new_sources = []
+                for source in entry["sources"]:
```
manasaV3, line 28: nit: Validate that this is not already in the expected format, so we can even run this on files that have already been formatted without worrying about it. (Not a blocker)

manasaV3 (approved):

ingestion_tools/scripts/refactor_annotation_sources.py
daniel-ji: Done!

PR #128, opened 2024-07-10
Update schema in preparation for dataset config validation

daniel-ji:
The second of four PRs, as part of dataset configuration file validation. To be merged second. Updates schema files with:
- `required` fields
- `recommended` fields
- `ifabsent` (default value) fields 
- `min`/`max` values (for numerical schema fields)
- new types (e.g., `FloatFormattedString`)
- new schema classes based on annotation.sources refactor
- removing no longer used schema classes
- some deprecated notes

Updates were based on `template.yaml`. Also improves / fixes `schema.py` schema generation. Metadocs/ folder is re-`make`-d. 

Files that were primarily changed (and are not code-gen'd files):
- `schema/v1.1.0/schema.py`
- `schema/v1.1.0/metadata.yaml`
- `schema/v1.1.0/common.yaml`

Note that `any_of` is not currently working for LinkML Pydantic code generation (see https://github.com/linkml/linkml/issues/1521). So some schema updates are commented out and waiting on it. Specifically, we cannot use the newly created types (like `FloatFormattedString`) accept both formatted strings and floats/ints for fields, so we plan on parsing the formatted string before running validation as a temporary workaround for now. 

Note that the Pydantic models were also generated with a temporary-hotfix that hasn't been yet implemented by linkml (see https://github.com/linkml/linkml/pull/2201). If the docs are regenerated without this hotfix, validation will be incorrect, but only more lax (required multivalued fields that are not provided in config files will not be correctly reported as errors). In the near future, I can look into creating a Python wrapper class / using jinja2 templates to extend the current LinkML release to provide a consistent fix.


schema/v1.1.0/common.yaml
```diff
@@ -270,28 +285,49 @@ slots:
...
       descriptive_name: volts
       symbol: V
   tiltseries_aligned_tiltseries_binning:
     range: integer
+    # TODO: uncomment after any_of is fixed https://github.com/linkml/linkml/issues/1521
+    # any_of:
+    #   - range: integer
+    #   - range: IntegerFormattedString
     description: Binning factor of the aligned tilt series
+    ifabsent: 'int(1)'
```
jgadling, line 301: I'm curious about this. Why do we need to specify `'int(1)'` instead of just `1` ?

jgadling (approved):
I can't speak to which specific metadata fields are being deprecated, but this looks sane to me.

schema/v1.1.0/common.yaml
daniel-ji: To prevent code injection I believe, see: https://github.com/linkml/linkml/blob/48241d48829a4311d3a9c96779cb241d7b829e56/linkml/utils/ifabsent_functions.py#L62-L122

PR #129, opened 2024-07-10
Add validation scripts for dataset config files

daniel-ji:
The third of four PRs, as part of dataset configuration validation. 

- `.pre-commit-config.yaml`: ignore LinkML code-gen'd Pydantic model used for validation
- `dataset_config_validate.py`: New file that runs validation by attempting to create a Pydantic model for each YAML, reporting any errors.
- `dataset_config_merge.py`: Modified to accept filename keywords to exclude files from running through validation. 
- `dataset_config.yaml`: LinkML schema to model the configuration files themselves, built off of the database schema (`metadata.yaml`, `common.yaml`, `metadata_materialized.yaml`). 
- `dataset_config_validate.yaml`: LinkML schema that wraps `dataset_config.yaml`, used to generate Pydantic models for validation. 
- `dataset_config_models.py`: Generated Pydantic model. 
- `Makefile`: Modified to generate validation Pydantic code as well. 

renamed 10011 to 10011_draft.yaml.


.pre-commit-config.yaml
```diff
@@ -10,8 +10,12 @@ repos:
...
   hooks:
   - id: ruff
-    # This file is codegen'd and non-compliant :'(
-    exclude: ^schema/v1.1.0/metadata_models.py
+    # These files are codegen'd and non-compliant :'(
+    exclude: |
+      (?x)(
+          ^schema/v1.1.0/metadata_models.py|
+          ^schema/v1.1.0/dataset_config_models.py
+      )
```
jgadling, line 18: TBH I'm on the fence about whether we should just run ruff on the codegen'd files after they get created so they follow our python spec as well. 🤷‍♀️

schema/v1.1.0/dataset_config.yaml
```diff
@@ -0,0 +1,405 @@
...
+
+  # ============================================================================
+  # Dataset config classes
+  # ============================================================================
+  AnnotationEntity:
+    description: An annotation entity.
+    attributes:
+      metadata:
+        description: The metadata for the annotation.
+        range: Annotation
```
jgadling, line 142: So cooool to see this actually working! ❤️

schema/v1.1.0/dataset_config.yaml
```diff
@@ -0,0 +1,405 @@
...
+        multivalued: true
+        range: string
+      tomogram:
+        description: Include or exclude tomograms.
+        multivalued: true
+        range: string
+      voxel_spacing:
+        description: Include or exclude voxel spacings.
+        multivalued: true
+        range: string
```
jgadling, line 116: Ok, technically not all parents are valid for all classes -- for example it makes no sense to filter a Run by Tomogram (because tomograms are children of runs). I think this is 100% fine for now, but if you have time to think about whether we should make the schema stricter, it would be more accurate.

jgadling (approved):
LGTM - we can continue some of these conversations though.

schema/v1.1.0/dataset_config.yaml
daniel-ji: Sounds good, I'll definitely follow up with you about this soon!

PR #132, opened 2024-07-12
Template, schema, and config updates

daniel-ji:
See https://github.com/chanzuckerberg/cryoet-data-portal/issues/867.


ingestion_tools/scripts/common/yaml_files.py
```diff
@@ -35,13 +36,18 @@ def get_yaml_config_files(
         if filename in EXCLUDE_LIST:
             continue
         if any(keyword in filename for keyword in exclude_keywords_list):
-            print(f"[INFO]: Excluding {file} because it contains an exclude keyword")
+            if verbose:
+                print(f"[INFO]: Excluding {file} because it contains an exclude keyword")
```
jgadling, line 40: nit: Typically we'd use the python logging library here (docs -- https://docs.python.org/3/library/logging.html). For example, we'd set the default log output level to `INFO` and every call to `logger.info()` would print a message to stdout. Output like the stuff above that we want to hide would probably be wrapped with `logger.debug()` and if the `verbose` flag is set, we'd tell the logging library to set the default log output level to `DEBUG`

schema/v1.1.0/metadata-docs/Tomogram.md
```diff
@@ -455,6 +489,10 @@ attributes:
     required: true
     inlined: true
     inlined_as_list: true
+    pattern: (^[ ]*\{[a-zA-Z0-9_-]+\}[ ]*$)|(^SART$)|(^FOURIER SPACE$)|(^SIRT$)|(^WBP$)|(^UNKNOWN$)
```
jgadling, line 492: Did we get agreement on WBP as short-hand for Weighted Back Projection? It seems like a pretty hard-to-understand acronym unless we include a glossary somewhere to explain what this means.

jgadling (changes requested):
I'd like to know more about what the plan is for WBP before merging

ingestion_tools/scripts/common/yaml_files.py
manasaV3: I think we address fix this before merging the code.

jgadling (approved):
Ok, we'll have to document the reconstruction type algorithms somewhere soon, but that can happen in another PR

schema/v1.1.0/common.yaml
```diff
@@ -722,6 +714,16 @@ enums:
...
   # ============================================================================
+  annotation_method_type_enum:
+    description: Describes how the annotations were generated.
+    permissible_values:
+      manual:
+        description: Annotations were generated manually.
+      automated:
+        description: Annotations were generated semi-automatically.
+      hybrid:
+        description: Annotations were generated automatically.
```
manasaV3, line 725: nit: I believe the descriptions need to be flipped.

schema/v1.1.0/common.yaml
```diff
@@ -526,15 +516,17 @@ slots:
...
     range: string
     description: Processing software used to derive the tomogram
     recommended: true
   tomogram_reconstruction_method:
-    range: string
+    any_of:
+      - range: StringFormattedString
+      - range: tomogrom_reconstruction_method_enum
     required: true
     description: Describe reconstruction method (Weighted back-projection, SART, SIRT)
```
manasaV3, line 531: nit: `Weighted back-projection` -> `WBP`

schema/v1.1.0/metadata-docs/Tomogram.md
manasaV3: The `WBP` was recommended by @uermel . But, I agree that we should add a glossary to explain all the acronyms for tomogram reconstruction_method. 

cc: @junxini  @dgmccart

schema/v1.1.0/common.yaml
```diff
@@ -785,11 +828,63 @@ types:
...
+
+  DOI:
+    description: A Digital Object Identifier
+    base: str
+    pattern: '^(doi:|https://doi\.org/)?10\.[0-9]{4,9}/[-._;()/:a-zA-Z0-9]+$'
+
+  DOI_LIST:
+    description: A list of Digital Object Identifiers
+    base: str
+    pattern: '^(doi:|https://doi\.org/)?10\.[0-9]{4,9}/[-._;()/:a-zA-Z0-9]+(\s*,\s*(doi:|https://doi\.org/)?10\.[0-9]{4,9}/[-._;()/:a-zA-Z0-9]+)*$'
```
manasaV3, line 872: We should only allow the doi ids and not allow the links, as the frontend converts these ids to links.

daniel-ji: Fixed!

PR #134, opened 2024-07-12
Update schema to create files that don't cause conflict in names

manasaV3:
## Description

When a slot and class are a single word and have the same names, it creates files that have same names but are in different cases. This can be an issue for filesystems that are case insensitive. So, making updates to handle that. 

### Updates: 
`Camera` -> `CameraDetails`
`Funding` -> `FundingDetails`
`Microscope` -> `MicroscopeDetails`
`Tissue` -> `TissueDetails`
`Organism` -> `OrganismDetails`


daniel-ji (approved):
Fix looks good! It's worth to note that currently, we only have to be aware of these case-insensitive class-slot conflicts for metadata.yaml, but if we ever start putting classes in common.yaml (unlikely), we'll also have to cross-check there. It's maybe also worth noting that slots in common.yaml have the same name as the class attributes (under the hood, slots) they are being exactly mapped to (in metadata.yaml), but because they are both slots, this should be fine.

PR #135, opened 2024-07-12
Adds deposition entity to configs

manasaV3:
Relates to: https://github.com/chanzuckerberg/cryoet-data-portal/issues/783 

## Description
Updates all the existing config files with a deposition entity. 

### Todo
- [ ] Add valid information for all the depositions such as title, description


ingestion_tools/dataset_configs/10427.yaml
```diff
@@ -12,65 +12,237 @@ datasets:
...
+    deposition_title: TBA
+    deposition_type:
+    - dataset
+  sources:
+  - literal:
+      value:
+      - 10306
+frames:
+- sources:
+  - parent_filters:
```
jgadling, line 78: I think most of these sources have a lot in common... does the converter script need to be smart enough to just smash them all into a single source?

ingestion_tools/dataset_configs/10301.yaml
```diff
@@ -270,7 +283,7 @@ runs:
       match_regex: /.*_BrnoKrios_.*$
       name_regex: (.*)
 standardization_config:
-  deposition_id: 10301
+  deposition_id: 10300
```
jgadling, line 286: We can probably do away with this field now?

ingestion_tools/scripts/transform_ingestion_configs.py
```diff
@@ -9,6 +9,32 @@ def cli():
...
+        "dates": metadata.get("dates", {}),
+        "authors": metadata.get("authors", []),
+    }
+
+
+def has_no_sources(data: list[dict[str, Any]] | dict[str, Any]) -> bool:
+    return isinstance(data, dict) or not any(row.get("sources") for row in data)
+
+
 def update_config(data: dict[str, Any]) -> dict[str, Any]:
```
jgadling, line 38: I think a lot of the code in this file is moot now since we've already translated old-style configs to new-style configs. 😓

ingestion_tools/scripts/transform_ingestion_configs.py
```diff
@@ -146,10 +183,10 @@ def update_config(data: dict[str, Any]) -> dict[str, Any]:
                 ],
             },
         ]
-    if not data.get("voxel_spacing"):
+    if not data.get("voxel_spacings"):
         vs = data["tomograms"][0]["metadata"]["voxel_spacing"]
         # Make sure voxel spacing is a float.
-        if isinstance(vs, str) and "{" in vs:
+        if isinstance(vs, str) and "{" in vs and not vs.strip().startswith("float"):
```
jgadling, line 189: 👍

jgadling (approved):
It's a little hard to tell the *data* changes from the *format* changes in this PR - a bunch of the gjensen datasets have funding/author/etc changes alongside the deposition changes. I'm guessing it was just easiest to pull in both sets of changes.

ingestion_tools/dataset_configs/10301.yaml
manasaV3: I can remove this, after I update the ingestion PR to source the deposition_id from the parent instead of standardization_config.

ingestion_tools/scripts/transform_ingestion_configs.py
manasaV3: Gjensen configs still do the old style generation, and are updated to the new style with this script.

jgadling: ahhh ok. Maybe I'll fix the gjensen config script someday.

manasaV3:
All updates to yaml files in this PR are formatting changes and addition of deposition. The only metadata changes for Grant Jensen dataset was part of a csv file, which has been reverted as it is unrelated to this PR. :)

daniel-ji (approved):
Looks good! Just to note that right now, the reconstruction_method is still creating "Weighted back projection", when validation / schema / template reflects that it expects an enum value of "WBP". But as Jess mentioned, this can definitely be another PR.

PR #137, opened 2024-07-15
dataset config files pydantic validation setup and annotations

daniel-ji:
See https://github.com/chanzuckerberg/cryoet-data-portal/issues/869. To merge **after** #136 (https://github.com/chanzuckerberg/cryoet-data-portal/issues/878)

Note that #140 fixes some caching bugs that are in here. (and also improves logging on validation pieces that were built here)


jgadling (approved):

schema/v1.1.0/dataset_config_models_extended.py
```diff
@@ -0,0 +1,353 @@
...
+running_publication_list: Dict[str, bool] = {}
+
+running_network_validation = False
+
+# ==============================================================================
+# Annotation Object Validation
+# ==============================================================================
+
+
+def validate_go_id(go_id: str) -> bool:
```
manasaV3, line 35: We could use the `@cache` here instead of maintaining a running list. 

ref: https://docs.python.org/3/library/functools.html

daniel-ji: Done!

.vscode/settings.json
```diff
@@ -0,0 +1,9 @@
+{
```
manasaV3, line 1: Wasn't this a part of a different PR? 🤓

.github/workflows/dataset-config-validate.yaml
```diff
@@ -30,18 +28,22 @@ jobs:
...
+          poetry config virtualenvs.create false
+          poetry install
+        working-directory: schema/
       - name: Build Pydantic models
         run: make build
         working-directory: schema/v1.1.0
       - name: Validate dataset configs
         run: |
-          python dataset_config_validate.py
+          make validate-configs-with-network
```
manasaV3, line 42: Yay for the abstraction!

.vscode/settings.json
daniel-ji: yes, sorry about this, must have messed up somewhere when merging branches

.github/workflows/dataset-config-validate.yaml
daniel-ji: it's lit, no cap!

ingestion_tools/scripts/gjensen_config.py
```diff
@@ -105,6 +105,9 @@ def to_dataset_config(
         config["tissue"] = dataset["tissue"]
     if cross_reference:
         config["cross_references"] = cross_reference
+        if "dataset_publications" in config["cross_references"]:
+            config["cross_references"]["publications"] = config["cross_references"]["dataset_publications"]
+            del config["cross_references"]["dataset_publications"]
```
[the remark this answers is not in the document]
daniel-ji: Validation doesn't support dataset_publications, so we should probably keep it?

manasaV3 (approved):

.vscode/settings.json
manasaV3: All good!

.github/workflows/dataset-config-validate.yaml
manasaV3: no cap!

ingestion_tools/scripts/gjensen_config.py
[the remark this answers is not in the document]
manasaV3: Because of the changes [here](https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/db8268874e758627a56e09029226be38c71da296/ingestion_tools/scripts/gjensen_config.py#L385-L394), we might no longer need this.

manasaV3 (approved):

ingestion_tools/dataset_configs/gjensen/10172.yaml
daniel-ji, line 1: Note that although the original "endothelial cell of umbilical vein" is the valid name (it passes validation checks), it gets overwritten by the invalid version "umbilical vein endothelial cell" (the proposed change). Why are we overwriting the correct version with the wrong version? Because this should not be an issue fixed here, but the fix should be in the data files that gjensen_config.py pulls from so that a re-run of gjensen_config.py doesn't later override the correct values with a stored incorrect version.

ingestion_tools/dataset_configs/gjensen/10162.yaml
```diff
@@ -18,7 +18,7 @@ datasets:
       id: null
       name: ATCC 15444
     cross_references:
-      publications: https://doi.org/10.1111/mmi.13593
+      publications: 10.1111/mmi.13593
```
daniel-ji, line 21: These https://doi.org/ prefixes are replaced with a re-run of the proposed new gjensen_config.py changes

ingestion_tools/pyproject.toml
```diff
@@ -46,8 +46,7 @@ boto3-stubs = {extras = ["s3"], version = "^1.34.34"}
 mypy = "^1.8.0"
 cryoet-data-portal-neuroglancer = { git = "https://github.com/chanzuckerberg/cryoet-data-portal-neuroglancer.git", branch = "main" }
 distinctipy = "^1.3.4"
-imodmodel = ">=0.0.12"
-linkml = "^1.8.1"
+imodmodel = "0.0.12"
```
daniel-ji, line 49: imodmodel = "0.0.13" has breaking changes, (see the rename of slan to slicer_angle), so as per @jgadling, we will stick to 0.0.12 for now. 

See https://github.com/teamtomo/imodmodel/commit/5908efc1993ff076bc6196ac992a5bf20476c4c3#diff-5073310c046e4db3049d5d7f98bf4fa26775dd55aab3ec4c4d7e4defb780178e

schema/v1.1.0/dataset_config_models_extended.py
```diff
@@ -0,0 +1,321 @@
+from __future__ import annotations
```
daniel-ji, line 1: A non-auto generated file. This is where the extended validation (validation that linkml can't do) is written, based closely off the classes of `dataset_config_models.py`

schema/v1.1.0/dataset_config_models_extended.py
```diff
@@ -0,0 +1,321 @@
...
+# ==============================================================================
+# Annotation Object Validation
+# ==============================================================================
+
+
+@alru_cache
+async def validate_go_id(go_id: str) -> Tuple[bool, dict]:
+    url = f"https://api.geneontology.org/api/ontology/term/{go_id}"
+    headers = {"Accept": "application/json"}
+    async with aiohttp.ClientSession() as session, session.get(url, headers=headers) as response:
```
daniel-ji, line 34: creating a aiohttp session has negligible overhead, so it's fine to create one everytime

schema/v1.1.0/dataset_config_models_extended.py
```diff
@@ -0,0 +1,321 @@
...
+@alru_cache
+async def is_go_id_ancestor(go_id_ancestor: str, go_id: str) -> bool:
+    url = f"https://api.geneontology.org/api/ontology/term/{go_id}/subgraph?start=0&rows=100"
+    headers = {"Accept": "application/json"}
+    async with aiohttp.ClientSession() as session, session.get(url, headers=headers) as response:
+        response_json = await response.json()
+        return go_id_ancestor in [ancestor["id"] for ancestor in response_json["ancestors"]]
+
+
+class ExtendedValidationAnnotationObject(AnnotationObject):
```
daniel-ji, line 48: classes are getting extended so that their existing validators built by linkml are preserved. note however that if fields are not copied over, weird behavior happens in some odd cases, so in the code following, the field values are re-assigned their value (in addition changing the field type to reference an extended-validation version of the original class)

schema/v1.1.0/dataset_config_models_extended.py
```diff
@@ -0,0 +1,321 @@
...
+    async with aiohttp.ClientSession() as session, session.head(url) as response:
+        return emdb_id, response.status == 200
+
+
+PUBLICATION_REGEXES_AND_FUNCTIONS = {
+    "doi": (r"^(doi:)?10\.[0-9]{4,9}/[-._;()/:a-zA-Z0-9]+$", lookup_doi),
+    "empiar": (r"^EMPIAR-[0-9]{5}$", lookup_empiar),
+    "emdb": (r"^EMD-[0-9]{4,5}$", lookup_emdb),
+}
+
```
daniel-ji, line 110: ^ is created for easier handling of the different types of publications and their according regexes / lookup functions

(if a list entry in a publication list matches a regex, then run the according lookup function to verify

schema/v1.1.0/dataset_config_models_extended.py
```diff
@@ -0,0 +1,321 @@
...
+        if author.primary_author_status:
+            primary_author_status_count += 1
+        if author.corresponding_author_status:
+            corresponding_author_status_count += 1
+
+    if primary_author_status_count == 0:
+        errors.append(ValueError("Annotation must have at least one primary author"))
+    if corresponding_author_status_count == 0:
+        errors.append(ValueError("Annotation must have at least one corresponding author"))
+
```
daniel-ji, line 158: doing this so that both ValueErrors can be returned if both fail (so one ValueError doesn't short-circuit the function and prevent the other one from getting checked)

PR #144, opened 2024-07-17
Schema changes: Refactor cross_references.dataset_publications -> cross_references.publications, add PDB as database entry

daniel-ji:
Refactor cross_references.dataset_publications -> cross_references.publications in preparation of also having cross_references as an entry for depositions. Also add PDB as a valid value for cross_references.related_database_entries and annotations.annotation_publications:

Note: annotations still has an entry of annotation_publications, but in the future, we would likely want to modify the database so that annotations also has a cross_references? 

Thanks!


jgadling (approved):

ingestion_tools/scripts/gjensen_config.py
```diff
@@ -105,6 +105,9 @@ def to_dataset_config(
         config["tissue"] = dataset["tissue"]
     if cross_reference:
         config["cross_references"] = cross_reference
+        if "dataset_publications" in config["cross_references"]:
```
manasaV3, line 108: What are your thoughts on moving this clean up here: https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/db79d2e69dc94a4bc6ac12d8583df673f5f5d518/ingestion_tools/scripts/gjensen_config.py#L383?

ingestion_tools/scripts/importers/db/dataset.py
```diff
@@ -42,7 +42,7 @@ def get_direct_mapped_fields(cls) -> dict[str, Any]:
             "last_modified_date": ["dates", "last_modified_date"],
             "related_database_entries": ["cross_references", "related_database_entries"],
             "related_database_links": ["cross_references", "related_database_links"],
-            "dataset_publications": ["cross_references", "dataset_publications"],
```
manasaV3, line 45: Just as a safety measure, we should add a fallback here to check the old path if this is none.

ingestion_tools/dataset_configs/template.yaml
```diff
@@ -11,7 +11,7 @@ annotations: OPTIONAL
       id: REQUIRED, STRING (GO_ID)
       name: REQUIRED, STRING
       state: OPTIONAL, STRING
-    annotation_publications: OPTIONAL, STRING
```
manasaV3, line 14: This should also be broken down into cross_references.

manasaV3 (approved):

PR #145, opened 2024-07-17
Refactor Weighted back projection -> WBP, remove "None" instances

daniel-ji:
Refactored "weighted back projection" instances to WBP, as per template.yaml spec

Also removed some 'None' instances from tiltseries.phase_plate and tiltseries.image_corrector in the test SQL (the values could be NULL, so replaced them with NULL)

And per a previous conversation with Utz, we don't want "None" values, so they are now "UNKNOWN" values, as per template.yaml tomograms.reconstruction_method spec (and just did the same for tomograms.reconstruction_software, although they can be any string for reconstruction software).


jgadling (approved):

ingestion_tools/scripts/common/normalize_fields.py
```diff
@@ -11,5 +11,5 @@ def normalize_fiducial_alignment(status: bool | str) -> str:
     raise Exception("Fiducial alignment status must be FIDUCIAL or NON_FIDUCIAL")
 
 
-def normalize_to_none_str(value: str) -> str:
-    return value if value else "None"
+def normalize_to_unknown_str(value: str) -> str:
+    return value if value else "UNKNOWN"
```
manasaV3, line 15: Should this be "unknown"?

manasaV3 (approved):

PR #147, opened 2024-07-17
Change cell_strain and cell_type of 10426-10435 to conform to new validation rules

uermel:
## Description

Changing the values to have a valid cell ontology term in cell_type and a cell line ontology term in cell_strain.


daniel-ji (approved):
Looks good to me!

PR #149, opened 2024-07-18
pydantic validation: `pydantic__no_name_validation` flag

daniel-ji:
To be PR'd after #140.

Sometimes it doesn't make sense to check the ontology object's ID against the name and the name is intentionally different from what is stored online (this is for fields like annotation_object, cell_type, cell_component, tissue, cell_strain, etc.). With this PR, a new attribute can now be specified `pydantic__no_name_validation` (alongside the `name` and `id` attributes in the config file) to stop the Pydantic extended network validation from doing this check and throwing an error.


schema/v1.1.0/dataset_config_models_extended.py
```diff
@@ -0,0 +1,651 @@
...
+        return orcid_id, response.status == 200
+
+
+async def validate_orcids(orcid_list: List[str]) -> List[str]:
+    """
+    Returns a list of invalid ORCIDs, from the provided list
+    """
+    invalid_orcids: List[str] = []
+
+    tasks = [lookup_orcid(orcid) for orcid in orcid_list]
```
jgadling, line 86: 🫶

ingestion_tools/dataset_configs/template.yaml
```diff
@@ -86,18 +88,23 @@ datasets: REQUIRED
     cell_component: OPTIONAL
       id: OPTIONAL, STRING (GO_ID)
       name: REQUIRED, STRING
+      pydantic__no_name_validation: see annotations.annotation_object.pydantic__no_name_validation
```
jgadling, line 91: This seems a little dirty to me that we're adding a field to our **ingestion schema** that toggles ontology verification on/off. What's the advantage to doing it here vs making it a runtime flag like `--enforce-ontologies` to the validation script? Are we trying to enforce ontologies by default and just recording this escape hatch in our config files so it's explicit?

daniel-ji: I think the idea is that with a flag here, then individual entries can be marked to skip validation, as if we did --enforce-ontologies we could do something similar by specifying a file, but that may not give the same level of granularity.

daniel-ji:
We'll not use this method and instead store a json file that gives a whitelist of fields & corresponding values to not check for validation.

PR #150, opened 2024-07-18
Rename cell type, other dataset config files fixes

daniel-ji:
cell_type.name umbilical vein endothelial cell -> endothelial cell of umbilical vein to match what's stored online (and pass validation, since there's no fuzzy searching planned) see https://www.ebi.ac.uk/ols4/ontologies/cl/classes/http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FCL_0002618

Other various changes that make config files pass validation. 

Also updated validation to allow for lowercase letters in the ontology prefix.

### DIFF
--- a/ingestion_tools/dataset_configs/10001.yaml
+++ b/ingestion_tools/dataset_configs/10001.yaml
@@ -372,7 +372,7 @@ datasets:
       corresponding_author_status: true
       name: Judith B. Zaugg
     cell_strain:
-      id: https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?id=284812
+      id: NCBITaxon:284812
       name: K972 Sp h-wild type
     cross_references:
       publications: doi:10.1101/2022.04.12.488077, doi:10.1038/s41592-022-01746-2
--- a/ingestion_tools/dataset_configs/10008.yaml
+++ b/ingestion_tools/dataset_configs/10008.yaml
@@ -192,7 +192,7 @@ tiltseries:
       energy_filter: GIF Quantum LS
       phase_plate: null
     pixel_spacing: 3.52
-    related_empiar_entry: EMD-15053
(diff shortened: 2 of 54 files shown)


jgadling (approved):

ingestion_tools/dataset_configs/10008.yaml
```diff
@@ -192,7 +192,7 @@ tiltseries:
       energy_filter: GIF Quantum LS
       phase_plate: null
     pixel_spacing: 3.52
-    related_empiar_entry: EMD-15053
```
manasaV3, line 195: @uermel is there a reason we want this to capture just the EMPIAR values for tiltseries and not include the EMD entries?

ingestion_tools/dataset_configs/10001.yaml
```diff
@@ -372,7 +372,7 @@ datasets:
       corresponding_author_status: true
       name: Judith B. Zaugg
     cell_strain:
-      id: https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?id=284812
+      id: NCBITaxon:284812
```
manasaV3, line 375: We should co-ordinate with the FE engineers about this change, as they use this id for constructing the link. 
For example: https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?id=284812 has the relevant value we are looking for. But https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?id=NCBITaxon:284812  doesn't.

manasaV3 (approved):
LGTM, please link the issue to track the FE fix for the ontology update. :)

ingestion_tools/dataset_configs/10001.yaml
daniel-ji: Sounds good!

PR #152, opened 2024-07-18
add deposition_keyphotos to schema / metadata

daniel-ji:
A couple questions (related, but not blockers):

Since this reflects dataset_keyphotos, can either of these realistically have a parent_filters? Or anything other than a literal (source_glob, destination_glob, etc.)? Whatever the answers might be, I just wanted to also confirm these changes are reflected in dataset_key_photo.py. Thanks!

### DIFF
--- a/ingestion_tools/dataset_configs/10000.yaml
+++ b/ingestion_tools/dataset_configs/10000.yaml
@@ -372,7 +372,7 @@ datasets:
       corresponding_author_status: true
       name: Judith B. Zaugg
     cell_strain:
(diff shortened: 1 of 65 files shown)


jgadling (approved):

jgadling:
deposition key photos have a deposition parent, and dataset key photos have depositions and datasets as parents, so if we have a config that specifies multiple depositions or mutliple datasets, those filters would be valid.

In *practice* we're not really using these config files to represent multiple depositions, and if we're ingesting keyphotos, we're typically only ingesting a single dataset, but we should probably document that it's possible to use the filters.

manasaV3 (approved):

PR #153, opened 2024-07-19
Updating deposition ingestion

manasaV3:
Relates to https://github.com/chanzuckerberg/cryoet-data-portal/issues/784

## Description
Updates the deposition id to be sourced from the deposition parent entity instead of `standardized_config`. 

### TODO
- [ ] Delete the deposition_id entry in all the dataset config's standardized_config section


ingestion_tools/scripts/common/metadata.py
```diff
@@ -73,7 +73,7 @@ def get_filename_prefix(self, output_dir: str, identifier: int) -> str:
                 [
                     str(identifier),
                     re.sub("[^0-9a-z]", "_", obj.lower()),
-                    re.sub("[^0-9a-z.]", "_", f"{version.lower()}"),
+                    re.sub("[^0-9a-z.]", "_", str(version).lower()),
```
manasaV3, line 76: Updated, as the version is now set to be a float in the config.

jgadling (approved):

PR #155, opened 2024-07-19
869 pydantic validation: setup, annotations, datasets, depositions, tiltseries, tomograms, and sources

daniel-ji:
EDIT: A merged version of the three validation PRs (#137, #140, and this one).

See entire https://github.com/chanzuckerberg/cryoet-data-portal/issues/869 description.

Thanks!


ingestion_tools/dataset_configs/gjensen/10116.yaml
daniel-ji, line 1: Revert to czii data portal value (it's not valid, but we want to fix the bug in the source dataset there)

ingestion_tools/dataset_configs/gjensen/10162.yaml
daniel-ji, line 1: new gjensen_config.py script generates this fix

ingestion_tools/dataset_configs/gjensen/10172.yaml
daniel-ji, line 1: also reverting this back to incorrect value that is stored in czii repo, will fix there

ingestion_tools/pyproject.toml
daniel-ji, line 1: imodmodel = "0.0.13" has breaking changes, (see the rename of slan to slicer_angle), so as per @jgadling, we will stick to 0.0.12 for now. 

See https://github.com/teamtomo/imodmodel/commit/5908efc1993ff076bc6196ac992a5bf20476c4c3#diff-5073310c046e4db3049d5d7f98bf4fa26775dd55aab3ec4c4d7e4defb780178e

schema/v1.1.0/common.yaml
daniel-ji, line 1: better naming convention

schema/v1.1.0/dataset_config_models.py
daniel-ji, line 1: autogenerated

schema/v1.1.0/dataset_config_models.schema.json
daniel-ji, line 1: autogenerated

schema/v1.1.0/dataset_config_validate.py
daniel-ji, line 1: small script improvements and incorporating the new extended validation Pydantic model and a network validation flag

schema/v1.1.0/dataset_config_models_extended.py
```diff
@@ -0,0 +1,964 @@
...
+    return invalid_orcids
+
+
+def validate_authors(authors: List[Author]) -> List[ValueError]:
+    """
+    Returns a list of errors found in the authors list
+    """
+    all_errors = validate_authors_status(authors)
+
+    if running_network_validation:
```
manasaV3, line 151: nit: It is generally a good practice to keep the number of indents to be few, as that makes it easier to follow. While, this code functionally accurate, it would be even more easier to follow, if we updated this check to be 
```
if not running_network_validation:
    return all_errors
...
```

schema/v1.1.0/dataset_config_models_extended.py
```diff
@@ -0,0 +1,964 @@
...
+        orcids = list({author.ORCID for author in authors if author.ORCID is not None})
+
+        if len(orcids) == 0:
+            return all_errors
+
+        logger.debug("Checking ORCIDs: %s", orcids)
+        invalid_orcids = asyncio.run(validate_orcids(orcids))
```
manasaV3, line 158: nit: we don't have to cast the set as a list. We can set the type in the method called as an `Iterable[str]` object. Also, one nice thing with if conditions in python is that they evaluate to `False` if an iterable variable is empty.


```
orcids = {author.ORCID for author in authors if not author.ORCID }
if not orcids:
     return all_errors
```

schema/v1.1.0/dataset_config_models_extended.py
```diff
@@ -0,0 +1,964 @@
...
+    return invalid_publications
+
+
+def validate_publications(publications: Optional[str]) -> Optional[str]:
+    if not running_network_validation:
+        return publications
+
+    if publications is None or publications == "":
+        return publications
+
```
manasaV3, line 436: nit: This can be simplified to: 

```
if not running_network_validation or not publications:
    return publications
```

schema/v1.1.0/dataset_config_models_extended.py
```diff
@@ -0,0 +1,964 @@
...
+    # Convert the string to a list of publications
+    publication_list = publications.replace(" ", "").split(",")
+    if publication_list[-1] == "":
+        publication_list.pop()
```
manasaV3, line 440: nit:
```
publication_list = publications.replace(" ", "").rstrip(",").split(",")
```

schema/v1.1.0/dataset_config_models_extended.py
```diff
@@ -0,0 +1,964 @@
...
+                break
+
+    tasks = []
+    for publication_type, (_, validate_function) in PUBLICATION_REGEXES_AND_FUNCTIONS.items():
+        tasks += [validate_function(entry) for entry in list(new_publications[publication_type])]
+
+    results = await asyncio.gather(*tasks)
+    invalid_publications += [publication for publication, valid in results if not valid]
+
+    return invalid_publications
```
manasaV3, line 427: nit: could we simplify this as follows:

```
tasks = []
for publication in publication_list:
   for pub_type, (regex, validate_function) in PUBLICATION_REGEXES_AND_FUNCTIONS.items():
        if not re.match(regex, publication):
           continue
        updated_publication = publication.replace("doi:", "") if pub_type == "doi" else publication
        tasks.append(validate_function(updated_publication))
        break
results = await asyncio.gather(*tasks)
return [publication for publication, valid in results if not valid]        
```

manasaV3 (changes requested):

schema/v1.1.0/dataset_config_models_extended.py
```diff
@@ -0,0 +1,951 @@
...
+
+
+def validate_sources(
+    source_list: List[DefaultSource] | List[VoxelSpacingSource],
+    class_name: str,
+    skip_parent_filters: bool = False,
+) -> None:
+    total_errors = []
+
+    # For verifying that all source entries each only have one finder type
```
manasaV3, line 465: We should also validate that there is at least one source.

manasaV3 (approved):

PR #156, opened 2024-07-22
Represent current API/DB schema as a linkml document

jgadling:
This is a representation of our current API data model as a LinkML schema. It's here so that we can submit a further PR to define the changes we want to make to our schema for v2


schema/v1.1.0/common.yaml
```diff
@@ -755,6 +919,17 @@ enums:
...
+    description: Describes the shape of the annotation
+    permissible_values:
+      SegmentationMask:
+        description: A binary mask volume
+      OrientedPoint:
+        description: A series of coordinates and an orientation
+      Point:
+        description: A series of coordinates
+      InstanceSegmentation:
+        description: A volume with labels for multiple instances
```
daniel-ji, line 933: Should we have SemanticSegmentationMask as an entry here?

schema/v1.1.0/Makefile
daniel-ji, line 1: Should we also remove api_models_materialized.yaml when we run `make clean`?

schema/v1.1.0/common.yaml
```diff
@@ -741,6 +741,170 @@ slots:
...
+    required: true
+  api_https_prefix:
+    range: string
+    description: Path to a directory containing data for this entity as an HTTPS url
+    required: true
+  api_run_name:
+    range: string
+    description: Name of a run
+    required: true
+  api_tiltseries_spherical_aberration_constant:
```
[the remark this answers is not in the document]
jgadling: Yeah, that sounds cleanest tbh. I don't mind doing the update, but is this what you had in mind?

```
# metadata.yaml
      spherical_aberration_constant:
        exact_mappings:
        any_of:
          - range: cdp-common:tiltseries_spherical_aberration_constant
          - range: FloatFormattedString
```

schema/v1.1.0/common.yaml
jgadling: I don't think so! We support `SemanticSegmentationMask` as a `source` type in the ingestion tooling, but that gets written out as a standard `SegmentationMask` after ingestion:

https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/main/ingestion_tools/scripts/importers/annotation.py#L280-L282

schema/v1.1.0/common.yaml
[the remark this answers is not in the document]
daniel-ji: Yes, thank you!

schema/v1.1.0/api_models.yaml
```diff
@@ -0,0 +1,648 @@
...
+  # ============================================================================
+  # Mixins
+  # ============================================================================
+  IDMixin:
+    mixin: true
+    attributes:
+      id:
+        exact_mappings:
+          - cdp-common:api_sequential_identifier
+  DateStampedEntityMixin:
```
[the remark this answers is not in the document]
jgadling: Nope, I can take care of it!

schema/v1.1.0/schema.py
```diff
@@ -136,6 +136,7 @@ def _materialize_schema(schema: SchemaView, common_schema: SchemaView) -> Schema
                             slot["description"] = common_slot["description"]
                             slot["unit"] = common_slot["unit"]
                             a["range"] = common_slot["range"]
+                            range_type = a["range"]
```
jgadling, line 142: What's this for?

daniel-ji: In the case that a common_slot has a range_type that is an enum, we want to update the range_type variable so that the following if statement below gets hit and the slot's pattern gets updated with enum values. I can improve the clarity of this code, if you'd like, to something like:

```python
range_type = common_slot["range"]  # update range type to ensure enums are added in the next code block
```

or

```python
if a["range"] in schema_enums:
    for _, e in enumerate(schema.get_enum(a["range"]).permissible_values):
        slot["pattern"] = _add_to_slot_pattern(slot, f"^{e}$")
```

instead of `range_type = a["range"]`.

Thank you!

daniel-ji (approved):

PR #160, opened 2024-07-23
DRAFT: proposed changes for v2 schema

jgadling:
This is a draft of the schema changes that are coming for the v2 api

### DIFF
--- a/schema/Makefile
+++ b/schema/Makefile
@@ -41,9 +41,14 @@ clean-docs:
 
 .PHONY: build
 build: clean
-	python schema.py materialize $(METADATA_YAML) $(COMMON_YAML) $(METADATA_MATERIALIZED_YAML)
-	python schema.py materialize $(API_MODELS_YAML) $(COMMON_YAML) $(API_MODELS_MATERIALIZED_YAML)
-	python schema.py materialize $(INGESTION_CONFIG_MODELS_YAML) $(COMMON_YAML) $(INGESTION_CONFIG_MODELS_MATERIALIZED_YAML)
+	python3 schema.py materialize $(METADATA_YAML) $(COMMON_YAML) $(METADATA_MATERIALIZED_YAML)
+	python3 schema.py materialize $(API_MODELS_YAML) $(COMMON_YAML) $(API_MODELS_MATERIALIZED_YAML)
+	python3 schema.py materialize $(INGESTION_CONFIG_MODELS_YAML) $(COMMON_YAML) $(INGESTION_CONFIG_MODELS_MATERIALIZED_YAML)
+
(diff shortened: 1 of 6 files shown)


anchi2c (approved):

schema/v1.1.0/api_models.yaml
```diff
@@ -300,31 +380,216 @@ classes:
...
+      - IDMixin
+      - S3PrefixedEntityMixin
+    attributes:
+      deposition:
+        range: Deposition
+        inverse: Deposition.frames
+      per_section_parameters:
+        range: PerSectionParameters
+        inverse: PerSectionParameters.frames
+        multivalued: true
```
manasaV3, line 463: Should the relationship between the frame and per-section parameters be 1:1?

schema/v1.1.0/api_models.yaml
```diff
@@ -59,6 +59,67 @@ classes:
...
+      volume_z_offset:
+        exact_mappings:
+          - cdp-common:alignment_volume_z_offset
+      volume_x_rotation:
+        exact_mappings:
+          - cdp-common:alignment_volume_x_rotation
+      tilt_offset:
+        exact_mappings:
+          - cdp-common:alignment_tilt_offset
+      local_alignment_file:
```
manasaV3, line 118: I believe this is different from the actual alignment. If that is a correct assumption, should we also add fields to capture the file path to the alignment file?

schema/v1.1.0/api_models.yaml
anchi2c, line 1: I don't think these need to be in per-section alignment.  Let me consult Shawn Zheng at CZII tomorrow who develops alignment software.

schema/v1.1.0/api_models.yaml
anchi2c: each per-section parameters would map to a frame entry thru tiltseries and z-section

schema/v1.1.0/common.yaml
```diff
@@ -740,6 +743,139 @@ slots:
...
+      symbol: °
+
+  alignment_tilt_offset:
+    range: float
+    description: Additional tilt offset in degrees
+    unit:
+      descriptive_name: degrees
+      symbol: °
+
+  alignment_local_alignment_file:
```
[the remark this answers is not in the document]
jgadling: Honestly, I'm not sure if we even want this field at all - we decided to keep all of the alignment data directly in the db rather than relying on files in s3 for alignment data cc @uermel ?

schema/v1.1.0/api_models.yaml
jgadling: Frames can (at least technically) have multiple per-section-parameters, one for each tiltseries.

PR #164, opened 2024-07-24
Create validation exclusions for dataset config extended validation

daniel-ji:
Additional fixes: updated the dataset_config_validate_extended.py with correct `required` fields for `metadata` and root-level entries; cleaned up its python code as well

Adds a flag to enable ontology object id / name validation skipping (for fields annotation_object, cell_component, cell_strain, cell_type, organism, tissue). Addresses same concern as #149 (sometimes it doesn't make sense to check the ontology object's ID against the name and the name is intentionally different from what is stored online).

Feature description (also in schema/v1.1.0/docs/dataset_config_validate.yaml):

===================================

`--field-excludelist-file`
A JSON file specifying which class-field-value mappings do not need to be validated when running Pydantic **extended validation**. Note that requirement / pattern / enum / type validation will still be performed.

**Currently only supports skipping ontology object validation (dataset config fields: annotation_object, cell_component, cell_strain, cell_type, organism, tissue)**. This option is useful when you want to skip certain fields that intentionally fail validation. For example, sometimes validation doesn't want to be run on the name of a cell_strain, as it may not be what the cell strain's id corresponds to online.

JSON file format (note that the `ClassNameToSkipOn` is the class name in the `schema/v1.1.0/dataset_config_models.py`, which may be different from the class name in the dataset configuration file):
```json
{
    "ClassNameToSkipOn": {
        "field_name_to_skip_on1": ["field_value_to_skip1", "field_value_to_skip2"]
        "field_name_to_skip_on2": ["field_value_to_skip3", ...],
        ...
    },
    "AnotherClassNameToSkipOn": {
        ...
    },
    ...
}
```

Example file:
```json
{
    "AnnotationObject": {
        "id": ["GO:0030992", "GO:0035869"],
        "name": []
    },
    "CellType": {
        "id": [],
        "name": ["umbilical vein endothelial cell"]
    }
}
```


schema/v1.1.0/dataset_config_models_extended.py
```diff
@@ -84,6 +84,7 @@
 
 logger = logging.getLogger("dataset_config_validate")
 
+FIELD_EXCLUDELIST = {}
```
manasaV3, line 87: If a variable is named in all caps, we should treat it as a constant. 

Also, what is your thought on renaming this to `VALIDATION_EXCLUSIONS`?

schema/v1.1.0/dataset_config_models_extended.py
```diff
@@ -101,6 +102,33 @@
 running_network_validation = False
 
 
+# ==============================================================================
+# Helper Functions
+# ==============================================================================
+def check_excludelist_skip_validation(obj: BaseModel, field_name: str, case_sensitive: bool = False) -> bool:
```
manasaV3, line 108: nit: Thoughts on naming the methods `check_skip_validation`?

schema/v1.1.0/dataset_config_models_extended.py
daniel-ji: Sure, I'll go ahead and change this to `validation_exclusions`

schema/v1.1.0/dataset_config_models_extended.py
daniel-ji: will do!

manasaV3 (approved):
🚢

PR #168, opened 2024-07-25
update dataset config validation docs with poetry install requirements

daniel-ji:
Running dataset config validation requires scripts from the ingestion_tools and those scripts require their corresponding dependencies. A temporary workaround for properly setting up dataset config validation. A future goal could be to unify all potential pyproject.toml / poetry.lock files across the repo into one.


jgadling (approved):

daniel-ji:
Not doing this, instead removing schema/ dependency on ingestion_tools/

Issue #166, opened 2024-07-25
TomoConverter `update_headers` function throwing ValueError


daniel-ji:
For the following line in ingestion_tools/scripts/common/image.py, if the value of `self.volume_reader.get_mrc_extended_header()` is a numpy array, then an error will get thrown (see below):
```python
if ext := self.volume_reader.get_mrc_extended_header():
    mrcfile.set_extended_header(ext)
```

```
2024-07-25 03:45:41.068 wdl.w:cryoet_data_ingestion.t:call-cryoet_data_ingestion_workflow.stderr VERBOSE ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/5c96afb1dc885df6f6e546521dee7518cd2d41a5/ingestion_tools/scripts/common/image.py#L356

Thanks!

uermel:
Fixed with https://github.com/chanzuckerberg/cryoet-data-portal-backend/pull/169

PR #169, opened 2024-07-25
Fix setting extended header when importing MRC

uermel:
Relates to
#166

## Description
The property `MrcObject.extended_header` may return `None`, or values of type `np.recarray` or `np.ndarray`. In contrast to python lists, `np.ndarray` is not truthy, so comparison with any `np.ndarray` will raise exceptions. In order to circumvent, we ensure that the result of `MRCReader.get_mrc_extended_header()` is a np.array and then convert to a python list that enables truthy/falsy comparison. 

Covers the following cases: 
1. do set extended header on filled `np.ndarray` > `np.array(val: np.ndarray).tolist() -> list`
2. do set extended header on filled `np.recarray` > `np.array(val: np.recarray).tolist() -> list`
3. skip setting extended header on empty `np.ndarray` > `np.array(np.empty(0)).tolist() == []`
4. skip setting extended header on `None` > `np.array(None).tolist() == None`

### DIFF
--- a/ingestion_tools/scripts/common/image.py
+++ b/ingestion_tools/scripts/common/image.py
@@ -132,7 +132,7 @@ class VolumeReader(ABC):
     def get_volume_info(self) -> VolumeInfo:
         pass
 
-    def get_mrc_extended_header(self) -> np.record | None:
+    def get_mrc_extended_header(self) -> np.ndarray | None:
         return None
 
     def get_mrc_header(self) -> np.record | None:
@@ -161,7 +161,7 @@ class MRCReader(VolumeReader):
     def get_mrc_header(self) -> np.rec.array:
         return self.header
 
-    def get_mrc_extended_header(self) -> np.rec.array:
+    def get_mrc_extended_header(self) -> np.ndarray:
         return self.extended_header
 
     def get_pyramid_base_data(self) -> np.ndarray:
@@ -353,7 +353,10 @@ class TomoConverter:
             if old_header.exttyp.item().decode().strip():
                 header.nsymbt = old_header.nsymbt
                 header.exttyp = old_header.exttyp
-                if ext := self.volume_reader.get_mrc_extended_header():
+                ext = self.volume_reader.get_mrc_extended_header()
+                # Converting to list, so we can compare both for None and empty np.ndarray. Actually setting the
+                # ndarray.
+                if ext.tolist():
                     mrcfile.set_extended_header(ext)
 
         if header_mapper:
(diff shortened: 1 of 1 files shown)


ingestion_tools/scripts/common/image.py
```diff
@@ -353,7 +353,13 @@ def update_headers(self, mrcfile: MrcFile, header_mapper, voxel_spacing):
...
                 header.nsymbt = old_header.nsymbt
                 header.exttyp = old_header.exttyp
-                if ext := self.volume_reader.get_mrc_extended_header():
+                ext = self.volume_reader.get_mrc_extended_header()
+                # In order to cover the following cases:
+                # 1. do set on filled np.ndarray > np.array(val: np.ndarray).tolist() -> list
+                # 2. do set on filled np.recarray > np.array(val: np.recarray).tolist() -> list
+                # 3. skip setting on empty np.ndarray > np.array(np.empty(0)).tolist() == []
+                # 4. skip setting on None > np.array(None).tolist() == None
+                if np.array(ext).tolist():
```
manasaV3, line 362: Should the headers `exttyp` also be updated accordingly?

uermel: It is being updated in line 355, no?

manasaV3: Yes, but if we are not setting the extended header in some cases where there is an extended header in the original mrc file. Should that also reflect in the `exttyp` which is currently set to the value as original mrc file.

uermel: I see, so moving that inside of the if-clause would solve this issure, correct?

manasaV3: Yeah, I believe so. One quick sanity check, what would be the expected value for that header when there is no extended_header?

uermel: `header.exttyp = b'MRCO'`
`header.nsymbt = 0`

manasaV3: Sorry, in that case, should we reset only the `nsymbt` inside the if condition?

uermel: yep, I thought this has a default set somewhere else. I'll make the changes as requested.

uermel: On reflection, I think it's not necessary to set the extended header type, but we can set the extended header size to zero explicitely if there is no extended header.

manasaV3 (approved):

PR #170, opened 2024-07-25
fix: Segmentation mask output datatype to int

manasaV3:
Relates to

## Description
Fixes the issue of segmentation mask output being a float. The [converter selection](https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/66e0ddc0cffdafe8b0d71d678b7443a585337ca2/ingestion_tools/scripts/common/image.py#L418) is determined by if the label is `None`.

### DIFF
--- a/ingestion_tools/scripts/common/image.py
+++ b/ingestion_tools/scripts/common/image.py
@@ -270,7 +270,7 @@ class TomoConverter:
         # Ensure voxel spacing rounded to 3rd digit
         voxel_spacing = round(voxel_spacing, 3)
 
-        pyramid = [self.volume_reader.get_pyramid_base_data()]
+        pyramid = [self.scaled_data_transformation(self.volume_reader.get_pyramid_base_data())]
         pyramid_voxel_spacing = [(voxel_spacing, voxel_spacing, voxel_spacing)]
         z_scale = 2 if scale_z_axis else 1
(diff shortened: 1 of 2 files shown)


jgadling:
Note: at some point a test would be nice

manasaV3:
> Note: at some point a test would be nice

I agree. I think in addition to this being unit test, this is something we should add to our s3 validation. cc: @daniel-ji

uermel (changes requested):
This does not seem to fix the issue. To clarify, I am actually setting the mask_label in the config here: 

https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/f6e180988e670430768a6e0813f52b74acd6f53a/ingestion_tools/dataset_configs/10436.yaml#L103

uermel:
Found the bug: 

`TomoConverter.scaled_data_transformation` also has to be applied to the base layer of the pyramid here: https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/66e0ddc0cffdafe8b0d71d678b7443a585337ca2/ingestion_tools/scripts/common/image.py#L273

i.e. change to 

`pyramid = [self.scaled_data_transformation(self.volume_reader.get_pyramid_base_data())]`

uermel (approved):

PR #171, opened 2024-07-26
Fetch logs of enqueue_runs.py script

daniel-ji:
Creates a script that can retrieve log files of all jobs scheduled in an enqueue_runs.py run. Modifies enqueue_runs.py slightly to make this possible.


daniel-ji:
Note: this script has been helpful to Utz and definitely to me when we had to ingest data, definitely worthwhile to merge

PR #172, opened 2024-07-26
Refactor dataset config validation to support wormbase strains

daniel-ji:
Add both LinkML support and Pydantic network validation support.


ingestion_tools/scripts/common/yaml_files.py
daniel-ji, line 1: Not as related to supporting wormbase strain, but contains non-breaking refactor to cleanup the code and add input file support

schema/v1.1.0/dataset_config_models_extended.py
```diff
@@ -127,8 +127,8 @@ def check_skip_validation(obj: BaseModel, field_name: str, case_sensitive: bool
                 field_name,
                 field_value,
             )
-            return False
-    return True
+            return True
+    return False
```
daniel-ji, line 132: this function now returns whether or not to skip validation (makes more sense when reading it)

schema/v1.1.0/dataset_config_models_extended.py
```diff
@@ -227,9 +227,9 @@ def valid_dates(self) -> Self:
 # ID Object Validation
 # ==============================================================================
 @alru_cache
-async def validate_id(id: str) -> Tuple[dict, bool]:
+async def validate_id(id: str) -> Tuple[List[str], bool]:
```
daniel-ji, line 230: now returns synonyms and whether or not it is valid, not returning extra unnecessary data. this also makes it more modular, with the ability to swap this out for wormbase validation against a different api endpoint

ingestion_tools/scripts/common/yaml_files.py
```diff
@@ -12,6 +12,7 @@
 
 
 def get_yaml_config_files(
+    input_files: list[str] = None,
     include_glob: str = None,
     exclude_keywords_list: list[str] = EXCLUDE_KEYWORDS_LIST,
     dataset_configs_dir: str = DATASET_CONFIGS_DIR,
```
jgadling, line 18: It's best to not use mutable default values in python function definitions. If you're *very* diligent to never mutate those values it's technically safe, but imop it's easier to avoid the situation entirely:
https://docs.python-guide.org/writing/gotchas/

jgadling (approved):

ingestion_tools/scripts/common/yaml_files.py
daniel-ji: Adjusted!

schema/v1.1.0/dataset_config_models_extended.py
```diff
@@ -268,36 +274,75 @@ async def is_id_ancestor(id_ancestor: str, id: str) -> bool:
...
+
+    logger.debug("Getting ID %s at %s", id, url)
+
+    names = []
+    async with aiohttp.ClientSession() as session, session.get(url) as response:
+        if response.status >= 400:
+            return [], False
+        data = await response.json()
+        if label := data.get("name", {}).get("data", {}).get("label", ""):
+            names.append(label.lower())
```
manasaV3, line 293: Did we want the names to be case insensitive?

daniel-ji: Resolved!

PR #176, opened 2024-07-26
Decouple schema python scripts from ingestion_tools dependencies

daniel-ji:
Instead of importing `ingestion_tools` scripts with dependencies not installed by the schema pyproject.toml (in particular standardize_dirs.py and its `IMPORTER_DEP_TREE` variable), we now generate a json file based on it to use for dataset config validation (which is contained in the `schema` folder). Note that this generation is run manually.


schema/v1.1.0/dataset_config_models_extended.py
```diff
@@ -93,10 +77,9 @@
     ("FEI", "TFS"): ["FALCON IV", "Falcon4i"],
     ("Gatan", "GATAN"): ["K2", "K2 SUMMIT", "K3", "K3 BIOQUANTUM", "UltraCam", "UltraScan"],
 }
-FLATTENED_DEP_TREE = {
-    key.type_key: {value.type_key for value in value_set}
-    for key, value_set in flatten_dependency_tree(IMPORTER_DEP_TREE).items()
-}
+VALID_PARENT_FILTERS = None
+with open("valid_parent_filters.json", "r") as f:
+    VALID_PARENT_FILTERS = json.load(f)
```
jgadling, line 82: So we're basically extending our auto-generated pydantic models to ensure that a dataset config doesn't specify invalid parent filters. This will work, but I'm wondering if a better solution would be to make the LinkML specs correct - to only accept certain parent filters for any given object type, so that our pydantic models get codegen'd with the right validations from the start.

daniel-ji: Sounds good, I'll make these adjustments!

jgadling (approved):

PR #187, opened 2024-08-06
Updating Jensen configs to include deposition metadata

manasaV3:

### DIFF
--- a/ingestion_tools/scripts/gjensen_config.py
+++ b/ingestion_tools/scripts/gjensen_config.py
@@ -33,13 +33,25 @@ RELATED_DATES = {
 }
 
 
+def get_json_data(input_dir: str, file_name: str) -> dict[str, Any]:
+    with open(os.path.join(input_dir, file_name), "r") as file:
+        data = json.load(file)
+    return data
+
+
 def to_author(authors_data: dict[str, Any], existing_author_obj: dict[str, Any] = None) -> list[dict[str, Any]]:
-    primary_author = set(authors_data["first_authors"])
+    """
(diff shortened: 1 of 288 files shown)


ingestion_tools/scripts/gjensen_config.py
```diff
@@ -382,9 +387,61 @@ def get_deposition_map(input_dir: str) -> dict[int, int]:
...
+            deposition_ids.add(deposition_id_mapping.get(ds_id))
+            ds_cross_reference = cross_reference_mapping.get(ds_id, {})
+            if publication := ds_cross_reference.get("publications"):
+                publications.update([p.strip() for p in publication.split(",")])
+            if related_database_entry := ds_cross_reference.get("related_database_entries"):
+                related_database_entries.update([r.strip() for r in related_database_entry.split(",")])
+
+        entry_metadata = entry.get("deposition", {})
+        if len(deposition_ids) != 1:
+            print(f"Multiple deposition ids found for {entry_metadata['deposition_title']}: {deposition_ids}")
```
daniel-ji, line 412: Maybe 

`print(f"Exactly one deposition id should exist for {entry_metadata['deposition_title']}: {deposition_ids}")` instead? And the ValueError could be "Exactly one deposition id should exist for a single deposition"? Thanks!

daniel-ji (approved):

ingestion_tools/scripts/gjensen_config.py
manasaV3: I believe the norm is to express what went wrong in an exception message.

PR #188, opened 2024-08-06
Update table and column descriptions

andy-sweet:
Relates to https://github.com/chanzuckerberg/cryoet-data-portal/issues/894

## Description

This adds a migration that alters table and column comments. Some of the older comments were incorrect, inconsistent, had typos, or were missing. The new comments mostly come from the manually maintained docstrings in the Python client from https://github.com/chanzuckerberg/cryoet-data-portal .

The motivation for doing this to ensure we have GraphQL scheme comments/descriptions, which can then be used to automatically generate correct docstrings for the Python client (as in https://github.com/chanzuckerberg/cryoet-data-portal/pull/910).

I have tested this locally in conjunction with https://github.com/chanzuckerberg/cryoet-data-portal/pull/910 , which now pulls in the updated comments when using the locally hosted GraphQL server.

### DIFF
--- a/api_server/migrations/cryoetdataportal/1722635643562_alter_comments/up.sql
+++ b/api_server/migrations/cryoetdataportal/1722635643562_alter_comments/up.sql
@@ -1,76 +1,57 @@
 comment on table "public"."datasets" is E'Metadata for a dataset';
--- comment on column "public"."datasets"."authors" is E'The authors of this dataset';
 comment on column "public"."datasets"."cell_component_id" is E'If the dataset focuses on a specific part of a cell, the subset is included here';
 comment on column "public"."datasets"."cell_strain_id" is E'Link to more information about the cell strain';
--- comment on column "public"."datasets"."funding_sources" is E'The funding sources of this dataset';
 comment on column "public"."datasets"."related_database_links" is E'If a CryoET dataset is also deposited into another database, e.g. EMPIAR, enter the database identifier here (e.g.https://www.ebi.ac.uk/empiar/EMPIAR-12345/). Use a comma to separate multiple links.';
--- comment on column "public"."datasets"."runs" is E'The experiment runs of this dataset';
 
 comment on table "public"."dataset_authors" is E'Metadata for authors of a dataset';
--- comment on column "public"."dataset_authors"."dataset" is E'The dataset this author corresponds to';
 comment on column "public"."dataset_authors"."dataset_id" is E'Numeric identifier for the dataset this author corresponds to';
 comment on column "public"."dataset_authors"."id" is E'A numeric identifier for this author';
 comment on column "public"."dataset_authors"."primary_author_status" is E'Indicating whether an author is the main person associated with the corresponding dataset';
 
 comment on table "public"."dataset_funding" is E'Metadata for a dataset''s funding sources';
--- comment on column "public"."dataset_funding"."dataset" is E'The dataset this funding source corresponds to';
 comment on column "public"."dataset_funding"."dataset_id" is E'Numeric identifier for the dataset this funding source corresponds to';
 comment on column "public"."dataset_funding"."id" is E'A numeric identifier for this funding record';
 
 comment on table "public"."runs" is E'Metadata for an experiment run';
--- comment on column "public"."runs"."dataset" is E'The dataset this run is a part of';
 comment on column "public"."runs"."https_prefix" is E'The HTTPS directory path where this dataset is contained';
 comment on column "public"."runs"."name" is E'Short name for this experiment run';
--- comment on column "public"."runs"."tiltseries" is E'The TiltSeries that correspond to this run';
--- comment on column "public"."runs"."tomogram_voxel_spacings" is E'The tomogram voxel spacings created from this run';
 
 comment on table "public"."tomogram_voxel_spacings" is E'Metadata for a set of tomograms and annotations of a given voxel spacing';
--- comment on column "public"."tomogram_voxel_spacings"."annotations" is E'The annotations associated with this voxel spacing';
 comment on column "public"."tomogram_voxel_spacings"."https_prefix" is E'The HTTPS directory path where this tomogram voxel spacing is contained';
 comment on column "public"."tomogram_voxel_spacings"."id" is E'Numeric identifier (May change!)';
--- comment on column "public"."tomogram_voxel_spacings"."run" is E'The run this tomogram voxel spacing is a part of';
 comment on column "public"."tomogram_voxel_spacings"."run_id" is E'The ID of the run this tomogram voxel spacing is a part of';
 comment on column "public"."tomogram_voxel_spacings"."s3_prefix" is E'The S3 public bucket path where this dataset is contained';
--- comment on column "public"."tomogram_voxel_spacings"."tomograms" is E'Tomograms of this voxel spacing';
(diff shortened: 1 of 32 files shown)


andy-sweet:
This is a draft PR because I am unsure how to successfully define a comment/description on the relationships.

Based on [the Hasura docs](https://hasura.io/docs/latest/api-reference/schema-metadata-api/relationship/#schema-metadata-create-array-relationship), it looks like provide a `comment` for the relationship in the table's metadata YAML file should work. I did this for a few files here to illustrate. But seems to be ignored with the default `"An array relationship"` or `"An object relationship"` being used instead. I also tried a few quick guesses to fix this (e.g. quoting the comment string, using `description` instead of `comment`), but with no success.

The [Hasura docs](https://hasura.io/docs/latest/api-reference/schema-metadata-api/relationship/#schema-metadata-set-relationship-comment) also mention `set_relationship_comment` but I'm unsure where I would define that to take an effect.

For local deployment, I am simply running `make clean` then `make init`.

The [Hasura docs](https://hasura.io/docs/latest/migrations-metadata-seeds/manage-migrations/#create-manual-migrations) also mention `hasura metadata [apply|reload]`, but I'm assuming those are used for live updates and should be handled by `make init`.

api_server/migrations/cryoetdataportal/1722635643562_alter_comments/up.sql
```diff
@@ -0,0 +1,74 @@
+comment on table "public"."datasets" is E'Metadata for a dataset';
+comment on column "public"."datasets"."cell_component_id" is E'If the dataset focuses on a specific part of a cell, the subset is included here';
+comment on column "public"."datasets"."cell_strain_id" is E'Link to more information about the cell strain';
+comment on column "public"."datasets"."related_database_links" is E'If a CryoET dataset is also deposited into another database, e.g. EMPIAR, enter the database identifier here (e.g.https://www.ebi.ac.uk/empiar/EMPIAR-12345/). Use a comma to separate multiple links.';
+
+comment on table "public"."dataset_authors" is E'Metadata for authors of a dataset';
+comment on column "public"."dataset_authors"."dataset_id" is E'Numeric identifier for the dataset this author corresponds to';
+comment on column "public"."dataset_authors"."id" is E'A numeric identifier for this author';
```
manasaV3, line 8: Could we add a "(May change!)" to this identifier? As this could potentially change, and we don't want the user to have this hardcoded.

api_server/migrations/cryoetdataportal/1722635643562_alter_comments/up.sql
```diff
@@ -0,0 +1,74 @@
...
+
+comment on table "public"."dataset_funding" is E'Metadata for a dataset''s funding sources';
+comment on column "public"."dataset_funding"."dataset_id" is E'Numeric identifier for the dataset this funding source corresponds to';
+comment on column "public"."dataset_funding"."id" is E'A numeric identifier for this funding record';
+
+comment on table "public"."runs" is E'Metadata for an experiment run';
+comment on column "public"."runs"."https_prefix" is E'The HTTPS directory path where this dataset is contained';
+comment on column "public"."runs"."name" is E'Short name for this experiment run';
+
+comment on table "public"."tomogram_voxel_spacings" is E'Metadata for a set of tomograms and annotations of a given voxel spacing';
```
manasaV3, line 19: What are your thoughts on "Metadata for a given voxel spacing related to tomograms/annotations"?

api_server/migrations/cryoetdataportal/1722635643562_alter_comments/up.sql
```diff
@@ -0,0 +1,74 @@
...
+comment on column "public"."tomogram_voxel_spacings"."https_prefix" is E'The HTTPS directory path where this tomogram voxel spacing is contained';
+comment on column "public"."tomogram_voxel_spacings"."id" is E'Numeric identifier (May change!)';
+comment on column "public"."tomogram_voxel_spacings"."run_id" is E'The ID of the run this tomogram voxel spacing is a part of';
+comment on column "public"."tomogram_voxel_spacings"."s3_prefix" is E'The S3 public bucket path where this dataset is contained';
```
manasaV3, line 23: *tomogram voxel spacing

manasaV3:
Currently hasura manages the migrations application. When we do a deploy to an environment, hasura should automatically apply the new migrations that have been added since the timestamp it was last updated. The unix timestamp here maps to the folder prefix.

manasaV3 (commented):
LGTM!

andy-sweet:
> This is a draft PR because I am unsure how to successfully define a comment/description on the relationships.

Thanks to @daniel-ji for finding a related existing issue: https://github.com/hasura/graphql-engine/issues/9600

Which suggests that Hasura does not currently use the various ways to encode a comment on a relationship, in the resulting GraphQL schema.

api_server/migrations/cryoetdataportal/1722635643562_alter_comments/up.sql
andy-sweet: Done - push incoming. Let me know if there are any other IDs this should apply to that were previously missed.

api_server/migrations/cryoetdataportal/1722635643562_alter_comments/up.sql
andy-sweet: Works for me. I just expanded the `/` to ` and `.

Push incoming.

api_server/migrations/cryoetdataportal/1722635643562_alter_comments/up.sql
andy-sweet: Good catch! Push incoming.

andy-sweet:
Since Hasura does not seem to use the relationship comments, I removed them for now. The workaround in https://github.com/chanzuckerberg/cryoet-data-portal/pull/910/ is to generate the comment from the types involved in the relationship.

andy-sweet:
@manasaV3 : Just wanted to check that there's nothing wrong with throwing all the comment updates into a single migration with the generic name of `alter_comments`.

api_server/migrations/cryoetdataportal/1722635643562_alter_comments/up.sql
manasaV3: All ids other than the dataset and deposition id could potentially change. Lemme double check if any other ids need that update.

andy-sweet: For `annotation_authors.id`, I see `this may change!` after getting the schema from the endpoint through introspection.

I did catch a missing `may change!` in `dataset_funding.id`, so added that.

andy-sweet:
@manasaV3 : this still needs an approval. Feel free to add line new comments or suggestions if there are more required additions or corrections.

api_server/migrations/cryoetdataportal/1722635643562_alter_comments/up.sql
manasaV3: Sorry, I was going by the comments on my local graphql. Which strangely has the comment "Numeric identifier for this annotation author (this may not change!)". Must have been a glitch on my end. 
Screenshot 2024-08-08 at 10 41 23 AM

manasaV3 (approved):

PR #189, opened 2024-08-06
Restructure schema folder and use separate versioning for core/, api/, and ingestion_config/ schemas

daniel-ji:
Note: maybe not worth merging until #160?

Refactored / restructured schema folder to as follows:

`schema/` folder contains a .version file that stores the latest versions of each schema: core, api, and ingestion_config (formally known as dataset_config). The three are separately versioned, and each have their own `codegen/` folder to store auto-generated code. See the comments in the files for more details about the changes. 

Did not commit the moved metadata-docs/, would that be desired? 

Note: I have renamed many of the schema-related folders and references to ingestion_config instead of dataset_config (because they are no longer "dataset" configuration files, but could also be deposition configuration files and whatnot) as per the suggestion of @jgadling, but have not yet done that for the ingestion_tools/dataset_config. I can do that in this PR or another one, whatever works best. 

Also updated linkml to 1.8.2 here, which has some bug fixes regarding linkml Pydantic and JSONSchema generation (https://github.com/linkml/linkml/releases/tag/v1.8.2, #2201, #2225, #2252)


.github/workflows/ingestion-config-validate.yaml
daniel-ji, line 1: When any confg files change in the schema folder, validation is re-run. Do we want to make it the latest version?

schema/Makefile
```diff
@@ -0,0 +1,63 @@
...
+	gen-pydantic --black $(METADATA_MATERIALIZED_YAML) > $(METADATA_MODELS_PY)
+	gen-pydantic --black $(INGESTION_CONFIG_MODELS_MATERIALIZED_YAML) > $(INGESTION_CONFIG_MODELS_PY)
+	gen-json-schema $(INGESTION_CONFIG_MODELS_MATERIALIZED_YAML) > $(INGESTION_CONFIG_JSON_SCHEMA)
+	cd ingestion_config && ln -s $(INGESTION_CONFIG_VERSION) latest && cd ..
+
+# `make build` must be run before `make build-docs`
+.PHONY: build-docs
+build-docs: clean-docs
+	gen-doc -d $(METADATA_DOCS) --no-mergeimports --hierarchical-class-view --stacktrace $(METADATA_MATERIALIZED_YAML)
+
```
daniel-ji, line 56: A separate set of make commands `make build-docs` and `make clean-docs` to manage generated doc files. Is this okay?

schema/Makefile
```diff
@@ -0,0 +1,63 @@
...
+.PHONY: validate-configs
+validate-configs:
+	python $(VALIDATE_CONFIGS)
+
+.PHONY: validate-configs-with-network
+validate-configs-with-network:
+	python $(VALIDATE_CONFIGS) --network-validation
```
daniel-ji, line 63: Do we want to also provide a `--verbose` option?

schema/Makefile
```diff
@@ -0,0 +1,63 @@
...
+
+.PHONY: build
+build: clean
+	python schema.py materialize $(METADATA_YAML) $(COMMON_YAML) $(METADATA_MATERIALIZED_YAML)
+	python schema.py materialize $(API_MODELS_YAML) $(COMMON_YAML) $(API_MODELS_MATERIALIZED_YAML)
+	python schema.py materialize $(INGESTION_CONFIG_MODELS_YAML) $(COMMON_YAML) $(INGESTION_CONFIG_MODELS_MATERIALIZED_YAML)
+	gen-pydantic --black $(METADATA_MATERIALIZED_YAML) > $(METADATA_MODELS_PY)
+	gen-pydantic --black $(INGESTION_CONFIG_MODELS_MATERIALIZED_YAML) > $(INGESTION_CONFIG_MODELS_PY)
+	gen-json-schema $(INGESTION_CONFIG_MODELS_MATERIALIZED_YAML) > $(INGESTION_CONFIG_JSON_SCHEMA)
+	cd ingestion_config && ln -s $(INGESTION_CONFIG_VERSION) latest && cd ..
```
daniel-ji, line 50: Is it fine to create a symbolic link from the latest version's folder to a folder with the actual name of "latest"? It makes accessing files from outside of this directory easier. Do we want to do this for all three folders (api/, core/, ingestion_config/) and not just ingestion_config?

schema/Makefile
```diff
@@ -0,0 +1,63 @@
+.DEFAULT_GOAL := build
+
+# include $CORE_VERSION, $API_VERSION, $INGESTION_CONFIG_VERSION
+include .version
+
```
daniel-ji, line 5: versions are pulled from `schema/.version` file

schema/schema.py
daniel-ji, line 1: The import statements in a non-materialized (non-autogen'd) yaml file are just copied over, which means that if it's a relative import (say a ingestion_config file importing over metadata_materialized.yaml in core/), and the directory changes (the newly materialized file is in codegen/), then the relative path will break. We fix that here.

schema/core/v1.1.0/codegen/metadata_models.py
daniel-ji, line 1: These changes are auto-generated by linkml 1.8.2, fixing some bugs

schema/Makefile
jgadling: I'm not sure - let's try it and see how it goes? We can adjust it, if it winds up being clunky.

schema/Makefile
jgadling: Yeah, I think those two targets are *probably* all we're going to need. We can add more later if more needs come up!

schema/Makefile
jgadling: I think we can run the command directly (outside of `make`) if we need to get creative with the arguments

schema/Makefile
jgadling: Why do we want to do this *and* have a symlink? I figure one or the other should be enough?

daniel-ji: @jgadling I guess if we have only the symlink, then we won't have anywhere where we explicitly store the version (I don't know if this a real concern we should be worried about). If we only have the .version, the referencing the latest version might be a bit more verbose (where we have to read & parse / load in the .version folder, as opposed to just referencing the latest folder). Thoughts? Thank you!

jgadling: Ahhh, I see. I agree that just setting the version as the real directory name might not be *super* explicit, but it seems good enough. If you want to keep the version file though, can you document the 2-step process for upgrading versions in the schema README?

jgadling (approved):
Once the version thing is resolved (either document how to work with symlinks/version files, or drop one or the other) this lgtm. Happy to see our schemas evolve independently!

schema/Makefile
daniel-ji: I can go ahead and update everything to use "latest" (so no version file), and then I'll document how to upgrade versions in the README (one-step process, just update the symlink). I'll also add a note how the latest version is just the "latest" folder, which if you really want to know which version that is, you can just `readlink` the symlink, which should anyways just be the highest-versioned folder. Does that sound okay?

daniel-ji: Now thinking about it, I think we should keep the .version number because ingestion_config/ is the only folder that needs a latest/ folder right now, so might just better to explicitly state it for everything and keep the latest/ folder for ingestion_config/, changing in the future if necessary. The extra step shouldn't be too much of a hassle and I've gone ahead and updated the README with instructions accordingly. Thanks!

PR #197, opened 2024-08-12
Titles and descriptions for annotation depositions.

uermel:

## Description

Deposition titles and descriptions for annotations depositions.

### DIFF
--- a/ingestion_tools/dataset_configs/deposition_10303_gas_vesicles.yaml
+++ b/ingestion_tools/dataset_configs/deposition_10303_gas_vesicles.yaml
@@ -39,9 +39,10 @@ depositions:
 - metadata:
     authors: *id001
     dates: *id002
-    deposition_description: TBA
+    deposition_description: Gas vesicle shell segmentations predicted using membrain-seg without rescaling and weights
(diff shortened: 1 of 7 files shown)


ingestion_tools/dataset_configs/deposition_10303_gas_vesicles.yaml
```diff
@@ -39,9 +39,10 @@ depositions:
 - metadata:
     authors: *id001
     dates: *id002
-    deposition_description: TBA
+    deposition_description: Gas vesicle shell segmentations predicted using membrain-seg without rescaling and weights
```
manasaV3, line 42: Both this and `deposition_10303.yaml` have the same deposition id. So they should have the same description and title.

uermel: Yeah, I guess ideally it would be two depositions (we created them before adding the additional metadata fields). Would that still be possible to do?

manasaV3: Yeah, we can do that.

manasaV3 (approved):

PR #198, opened 2024-08-12
Updating deposition id for gas vesicle to 10308

manasaV3:

## Description
Updates deposition id for gas vesicle to an unused deposition id.


uermel (approved):
LGTM!

PR #209, opened 2024-08-16
feat: ingest stl, vtk, and obj mesh annotations

Bento007:
Relates to https://github.com/chanzuckerberg/cryoet-data-portal/issues/951

## Description
This PR adds support to ingest mesh data as a new annotation type. These mesh formats are now supported stl, vtk, and obj mesh.

## Changes
- Add the `TriangularMeshAnnotation` class to support converting and uploading mesh annotation from a config file.'
- Add functions to convert multiple formats of mesh to the glb format.
- add trimesh as a dependency to support ingestion of stl, and obj mesh formats.
- add scale_factor and ontology_id as attributes in the ingestion config.
- update the schema with triangular mesh.
- add vtk as a dependency to support ingestion of vtk mesh formats.
- update poetry lock
- added test infrastructure to test on a local file system.


ingestion_tools/scripts/common/mesh_converter.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+    return trimesh.load(input_file)
+
+
+if __name__ == "__main__":
+    from_stl(
+        "/Users/trentsmith/workspace/cryoet/cryoet-data-portal-backend/test_infra/test_files/input_bucket/20002/annotations/Endospore.stl",
+        "/Users/trentsmith/workspace/cryoet/cryoet-data-portal-backend/test_infra/test_files/input_bucket/20002/annotations/Endospore.glb",
+        1.0,
+    )
```
manasaV3, line 106: nit: I believe these got retained by accident?

ingestion_tools/scripts/common/mesh_converter.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def convert_mesh_to_glb(convert_func: typing.Callable[[str], trimesh.Trimesh]) -> typing.Callable:
+    """
+    A decorator to convert and scale a mesh by a given factor
+
+    :param convert_func: the function to convert a mesh file to a trimesh object.
+    """
+
+    def wrapper(input_file: str, output_file: str, scale_factor: float = 1.0) -> None:
```
manasaV3, line 17: As the mesh is 3d, should the scale factor be a list of 3 values?

manasaV3 (changes requested):

ingestion_tools/scripts/common/mesh_converter.py
Bento007: It doesn't have to be if everything is scaled by the same amount for each direction. Is that something we need to support?

Bento007: @uermel what do you think?

ingestion_tools/scripts/common/mesh_converter.py
```diff
@@ -0,0 +1,106 @@
...
+@convert_mesh_to_glb
+def from_obj(input_file: str) -> trimesh.Trimesh:
+    """
+    Load a obj file into a trimesh object. More information about this format can be found
+    here: https://en.wikipedia.org/wiki/Wavefront_.obj_file
+
+    :param input_file: the path to an obj file.
+    :return: the mesh loaded into a trimesh object.
+    """
+    return trimesh.load(input_file)
```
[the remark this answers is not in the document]
Bento007: They could be the same method, but I kept them separate since the docstring contained different information.

ingestion_tools/scripts/tests/conftest.py
```diff
@@ -63,3 +63,9 @@ def s3_client(endpoint_url: str) -> S3Client:
         endpoint_url=endpoint_url,
         config=Config(signature_version="s3v4"),
     )
+
+
+@pytest.fixture
+def local_test_data_dir() -> str:
+    cd = os.path.dirname(__file__).split("/")[0:-3]
+    return os.path.join("/", *cd, "test_infra/test_files")
```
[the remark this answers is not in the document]
Bento007: Reading files from moto3 was causing my pycharm debug environment to hang. Reading from the local files system let me get around this. This is the same reason why i added the test_infra directory as a volume do the docker image since i'm running the tests in docker.

ingestion_tools/scripts/tests/s3_import/test_annotations.py
```diff
@@ -948,3 +963,83 @@ def test_ingest_instance_point_data(
 
         # Check id
         assert exp_point["instance_id"] == point["instance_id"], f"Incorrect id for {case['case']}"
+
```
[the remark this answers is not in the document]
Bento007: This tests file is already 1k lines. Splitting it would make it easier to find relevant tests.

ingestion_tools/scripts/common/mesh_converter.py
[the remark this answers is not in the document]
Bento007: I consolidated them into a single function

ingestion_tools/scripts/tests/s3_import/test_annotations.py
```diff
@@ -948,3 +963,83 @@ def test_ingest_instance_point_data(
...
+        "files": [
+            {
+                "format": "glb",
+                "path": "dataset1/run1/Tomograms/VoxelSpacing10.000/Annotations/100-some_protein-1.0-1_triangularmesh.glb",
+                "shape": "TriangularMesh",
+                "is_visualization_default": False,
+            },
+        ],
+    }
+    assert anno.local_metadata == expected_local_metadata
```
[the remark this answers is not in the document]
Bento007: Is [this](https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/867ce3856d0d6cbe72271f13b871d8a0768847e8/ingestion_tools/scripts/importers/annotation.py#L161-L161) value set in the AnnotationImporter class

ingestion_tools/scripts/tests/s3_import/test_annotations.py
```diff
@@ -948,3 +963,75 @@ def test_ingest_instance_point_data(
...
+
+    # load the new mesh file
+    anno_file = anno.get_output_filename(anno.get_output_path())
+    trimesh.load(anno_file)
```
manasaV3, line 1037: What is your thought on adding a regression check here by having an expected mesh, that we can compare to the output of this line?

ingestion_tools/dataset_configs/template.yaml
```diff
@@ -84,6 +84,15 @@ annotations: OPTIONAL
       mask_label: OPTIONAL, INTEGER (DEFAULT 1)
     parent_filters: see InstanceSegmentation.parent_filters
     exclude: SEE InstanceSegmentation.exclude
+  - TriangularMesh:
```
manasaV3, line 87: @uermel Can you confirm that `TriangularMesh` is okay to refer to this annotation shape?

uermel: ✅ fine to call it Triangular mesh.

ingestion_tools/scripts/common/mesh_converter.py
uermel: One scale factor is enough. I don't know of any case where anisotropic scaling would be applied in regular cryoET processing. We also don't do that for points.

ingestion_tools/dataset_configs/template.yaml
```diff
@@ -84,6 +84,15 @@ annotations: OPTIONAL
...
+      file_format: see InstanceSegmentation.file_format
+      glob_string: see InstanceSegmentation.glob_string
+      glob_strings: see InstanceSegmentation.glob_strings
+      is_visualization_default: see InstanceSegmentation.is_visualization_default
+      ontology_id: OPTIONAL, STRING (ONTOLOGY ID)
```
Bento007, line 92:
```suggestion
```

ingestion_tools/dataset_configs/template.yaml
```diff
@@ -84,6 +84,14 @@ annotations: OPTIONAL
       mask_label: OPTIONAL, INTEGER (DEFAULT 1)
     parent_filters: see InstanceSegmentation.parent_filters
     exclude: SEE InstanceSegmentation.exclude
+  - TriangularMesh:
+      file_format: see InstanceSegmentation.file_format
+      glob_string: see InstanceSegmentation.glob_string
+      glob_strings: see InstanceSegmentation.glob_strings
+      is_visualization_default: see InstanceSegmentation.is_visualization_default
+      scale_factor: OPTIONAL, FLOAT (DEFAULT 1) (POSITIVE)
```
manasaV3, line 92: Will the `ontology_id` field be added as part of the extension PR?

ingestion_tools/scripts/importers/annotation.py
```diff
@@ -476,3 +479,51 @@ def get_distinct_ids(self, output_prefix: str) -> set[int]:
...
+        output_file_name = self.get_output_filename(output_prefix, self.output_format)
+        tmp_path = self.config.fs.localwritable(output_file_name)
+        self.map_functions[self.file_format](mesh_file, tmp_path, scale_factor=self.scale_factor)
+        self.config.fs.push(tmp_path)
+
+    def get_object_count(self, output_prefix: str) -> int:
+        return 1
+
+    def is_valid(self) -> bool:
+        return True
```
manasaV3, line 529: nit: Is this a redundant?

schema/api/v1.0.0/codegen/ER_DIAGRAM.md
```diff
@@ -61,7 +61,7 @@ Annotation {
     EMPIAR_EMDB_DOI_PDB_LIST annotation_publication
     string annotation_method
     boolean ground_truth_status
-    GO_ID object_id
```
manasaV3, line 64: I believe this might be an unintended change?

ingestion_tools/dataset_configs/template.yaml
Bento007: Ontology id was a misnomer. It should be [biological annotation name](https://emdb-empiar.github.io/EMDB-SFF/v0.8.0.dev1/?_=x) which is just a name to describe the mesh as described in the EMDB-SFF spec. It is only needed for the hff file format which can contain multiple meshes per file. So it'll be added in the next PR.

Bento007: I plan to short hand the name to mesh_name in the future PR, that way it can be applied to future mesh formate that may support multiple meshes in a single file.

Should we enforce uniqueness among the `biological_annotation.names` within the hff?

manasaV3: We can add uniqueness of the `biological_annotation_names` to the config validation python extension.

ingestion_tools/scripts/tests/s3_import/test_annotations.py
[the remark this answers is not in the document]
manasaV3: Not sure, if you wanted to breakout the tests for meshes into its own file as a part of this.

manasaV3 (approved):

PR #211, opened 2024-08-19
Excluding runs from 10301 in 10302

manasaV3:
## Description
Fixes runs from dataset 10301 being included in dataset 10302 by explicitly excluding them.


ingestion_tools/dataset_configs/10302.yaml
```diff
@@ -96,6 +96,27 @@ rawtilts:
...
+        - ^15042022_BrnoKrios_Arctis_grid9_Position_32$
+        - ^15042022_BrnoKrios_Arctis_grid9_Position_65$
+        - ^17072022_BrnoKrios_Arctis_p3ar_grid_Position_101$
+        - ^17072022_BrnoKrios_Arctis_p3ar_grid_Position_35$
+        - ^17072022_BrnoKrios_Arctis_p3ar_grid_Position_68$
+        - ^17072022_BrnoKrios_Arctis_p3ar_grid_Position_76$
+        - ^27042022_BrnoKrios_Arctis_grid9_hGIS_Position_13$
+        - ^27042022_BrnoKrios_Arctis_grid9_hGIS_Position_33$
+        - ^27042022_BrnoKrios_Arctis_grid9_hGIS_Position_44$
+        - ^27042022_BrnoKrios_Arctis_grid9_hGIS_Position_7$
```
jgadling, line 119: If we exclude tomograms, rawtilts, etc from being imported by filtering out invalid runs from each sub-type, **we're still processing runs that are invalid for this dataset**. That's (barely) acceptable for the moment since we don't write any per-run data as output right now, but if we started writing `run_metadata.json` files, this falls apart. It seems like adding support for multiple include/exclude regexes to SourceGlobFinder would let us skip these runs at the appropriate level instead of repeating this list for multiple child types.

daniel-ji: So would an alternative (but likely hacky/clunky and non-sustainable) solution be adapting our current `runs.sources.source_glob` fields / regexes to use negative lookahead regex and filter out all of these?

jgadling: Yeah, in theory we could hack this right now with a super gnarly value in `runs.sources.source_glob.match_regex` value. It's probably not a ton of work to support multiple in/exclude regexes in that finder though 🤷‍♀️

manasaV3: I would lean towards the exclude over the regex as it is more cleaner with an exclude.

daniel-ji (approved):
Looks good to me!

ingestion_tools/scripts/common/finders.py
```diff
@@ -54,6 +56,8 @@ def __init__(
             name_regex = "(.*)"
         self.name_regex = re.compile(name_regex)
 
+        self.exclude_regexes = [re.compile(regex) for regex in exclude_regexes or []]
```
jgadling, line 59: It looks like we're supporting an `exclude_regexes` parameter for SourceGlobFinder as well as an `exclude` filter for any source type? It seems like the `exclude` filter would be enough and we don't need both?

jgadling (commented):
It looks like we're supporting an exclude_regexes parameter for SourceGlobFinder as well as an exclude filter for any source type? It seems like the exclude filter would be enough and we don't need both?

ingestion_tools/scripts/common/finders.py
manasaV3: Good catch! I initially added `exclude_regexes` to `SourceGlobFinder`. But, decided to generalize it to exclude for all the source types. Looks like the `exclude_regexes` got missed in clean up.

jgadling (approved):

PR #216, opened 2024-08-22
s3 data validation: rawtilts, runs

daniel-ji:
- Run metadata validation
- Rawtilt and tilt validation 
- Improve pytest performance for discovering run / voxel spacing folders

### DIFF
--- a/ingestion_tools/scripts/data_validation/conftest.py
+++ b/ingestion_tools/scripts/data_validation/conftest.py
@@ -1,3 +1,4 @@
+import concurrent.futures
 import os.path
 import sys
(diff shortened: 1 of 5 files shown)


ingestion_tools/scripts/data_validation/conftest.py
daniel-ji, line 1: This files' code changes is a performance improvement to retrieving which runs / voxel spacings to run tests on (so that the globs are more specific, especially when a run_glob is specified and we don't want to be globbing for voxel spacings in run folders that we're not even testing in)

PR #218, opened 2024-08-23
Fix comparison with mask_label during SemanticSegmentationMask ingestion

uermel:
## Description

As part of supporting Zarr-files in #75 the `TomoConverter.get_pyramid_base_data()` was replaced by `TomoConverter.volume_reader.get_pyramid_base_data()`. 

The old method used to be and should still be overridden in `MaskConverter.get_pyramid_base_data()` in order to convert a multi-label segmentation file to a binary segmentation file, wherever the image data is equal to `mask_label`. 

This PR re-introduces `TomoConverter.get_pyramid_base_data()` and uses `self.volume_reader.get_pyramid_base_data() ` in `MaskConverter.get_pyramid_base_data()` to access the data, should the input file be an old-style MRC-file with the wrong data type.


jgadling (approved):
A test to make sure this fix doesn't regress would be nice.

uermel:
> A test to make sure this fix doesn't regress would be nice.

Have added test now.

jgadling (approved):
thanks for the test!

manasaV3 (approved):

PR #220, opened 2024-08-23
ingestion config fixes

daniel-ji:
ingestion validation fixes & improvements

- better logging
- bugfix extended validation and add some missing checks
- add validation for new sources feature introduced in #211

### DIFF
--- a/schema/ingestion_config/v1.0.0/ingestion_config_models_extended.py
+++ b/schema/ingestion_config/v1.0.0/ingestion_config_models_extended.py
@@ -79,6 +79,7 @@ CAMERA_MANUFACTURER_TO_MODEL = {
     ("Gatan", "GATAN"): ["K2", "K2 SUMMIT", "K3", "K3 BIOQUANTUM", "UltraCam", "UltraScan"],
     ("simulated"): ["simulated"],
 }
+SOURCE_CHILDREN_NON_FINDERS = {"parent_filters", "exclude"}
 
 # Flag to determine if network validation should be run (set by provided Container arg)
 running_network_validation = False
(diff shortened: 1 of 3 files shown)


schema/ingestion_config/v1.0.0/ingestion_config_models_extended.py
```diff
@@ -651,6 +663,29 @@ def valid_sources(cls: Self, source_list: List[AnnotationSource]) -> List[Annota
...
+                    has_exclude = True
+                elif shape == "parent_filters":
+                    has_parent_filters = True
+                else:
+                    has_shape = True
+
+            if not has_shape and has_parent_filters:
+                only_parent_filter_in_source_entry_errors.append(i)
+            if not has_shape and has_exclude:
+                only_exclude_in_source_entry_errors.append(i)
```
manasaV3, line 687: Maybe we could have a constant for filter fields? 
Also, this could be combined with the iterations that are happening above? 
```suggestion
            for key_field in source_element.model_fields:
                if key_field not in {"parent_filters", "exclude"} and not getattr(source_element, shape):
                    continue            
                 has_shape = True
                 break

            if not has_shape: 
               total_errors.append(ValueError(f"Source entry {i} does not have a shape"))
               for filter_fields in {"parent_filters", "exclude"} :
                   if getattr(source_element, filter_fields):
                        total_errors.append(ValueError(f"Source entry {i} cannot have a {filter_fields} entry without a shape"))
```

schema/ingestion_config/v1.0.0/ingestion_config_models_extended.py
```diff
@@ -496,39 +501,46 @@ def validate_publications(publications: Optional[str]) -> Optional[str]:
...
+                has_exclude = True
+            elif finder == "parent_filters":
                 has_parent_filters = True
             else:
                 finders_in_source_entry.append(finder)
         # If there are more than one finder in the source entry, add the source entry index and the finders to the error set
         if len(finders_in_source_entry) > 1:
             multiple_finders_in_each_source_entries_errors.append((i, finders_in_source_entry))
-        elif len(finders_in_source_entry) == 0 and has_parent_filters:
+        if len(finders_in_source_entry) == 0 and has_parent_filters:
```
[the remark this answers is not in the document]
daniel-ji: Not necessarily, because parent_filters and exclude are considered entries in the default pydantic model. finders_in_source_entry tracks only actual "finders", e.g. source_glob, destinaton_glob, literal, etc (so when parent_filters / exclude exists, but actual finders don't, the length will still be 0). So in case where there's only parent_filters and/or exclude, it will pass default pydantic validation.

schema/ingestion_config/v1.0.0/ingestion_config_models_extended.py
daniel-ji: I was thinking that for the sake of clarity, we keep it separate?

manasaV3: Not sure, if this is making it more clearer. But, this is not a blocker.

daniel-ji: Ok, fixed!

schema/ingestion_config/v1.0.0/ingestion_config_models_extended.py
```diff
@@ -496,39 +502,29 @@ def validate_publications(publications: Optional[str]) -> Optional[str]:
...
-            # If the finder is not None, add it to the list of finders in the source entry
-            if getattr(source_element, finder) is None:
-                continue
-            if finder == "exclude":
-                continue
-            if finder == "parent_filters":
-                has_parent_filters = True
-            else:
+            # If the finder is not None and an actual finder, add it to the list of finders in the source entry
+            if getattr(source_element, finder) is not None and finder not in SOURCE_CHILDREN_NON_FINDERS:
```
manasaV3, line 511: nit: Could we rename this constant `SOURCE_FILTER_FIELDS` ?

daniel-ji: done!

manasaV3 (approved):

PR #222, opened 2024-08-23
fix: support ingesting volume annotations before point annotations

jgadling:
## Description
For annotations with both Point and non-point annotations, ingestion would fail if point files were ingested *after* non-point annotations. Basically, when we write metadata for an annotation, we expect to be able to get an object count from a point annotation file. If that file hasn't been written yet, ingestion fails.

This PR updates the metadata ingestion code to silently ignore missing points files - this is OK because we iterate over all of the annotation sources for a given annotation, and when we eventually write the points files, loading the number of points will succeed and we'll write the metadata file.


uermel (approved):
Looks good to me, ingestion worked with this fix applied.

PR #207, opened 2024-08-16
feat: s3 data validation: annotations, frames, gains

daniel-ji:
Creates s3 validation for annotations, frames, and gains. To be merged in after #228. See the annotation, frames, and gains sections of https://docs.google.com/document/d/1yMKM0DW9KRhlcYiBGPcR7oW0liGtUew6NAmBhMg5U3w/edit

I've left comments to describe some possibly confusing parts of the PR, but there are also some parts that I thought were relatively straight forward and left it out.

PR #223, opened 2024-08-25
feat: run / tiltseries s3 data validation (zarr, mrc, mdoc, and metadata checks)

daniel-ji:
Merge after #207.

See the tiltseries & runs sections of https://docs.google.com/document/d/1yMKM0DW9KRhlcYiBGPcR7oW0liGtUew6NAmBhMg5U3w/edit


ingestion_tools/scripts/data_validation/conftest.py
```diff
@@ -35,35 +36,56 @@ def run_names(bucket: str, dataset: str, run_glob: str) -> List[str]:
     return run_names
 
 
-def voxel_spacings(bucket: str, dataset: str, voxelspacing_glob: str) -> List[float]:
-    """All voxel spacings matching the vozel spacing glob pattern in the dataset."""
+def get_voxel_spacing_files(bucket: str, dataset: str, run_names: List[str], voxelspacing_glob: str) -> List[str]:
     fs: S3Filesystem = FileSystemApi.get_fs_api(mode="s3", force_overwrite=False)
```
manasaV3, line 40: nit: Why is this not a parameter?

ingestion_tools/scripts/data_validation/conftest.py
```diff
@@ -35,35 +36,56 @@ def run_names(bucket: str, dataset: str, run_glob: str) -> List[str]:
...
+        )
+
+        # Combine results
+        for result in results:
+            tentatives += result
 
     exclude = ["Images", "dataset_metadata.json"]
 
     for ex in exclude:
         tentatives = [tent for tent in tentatives if ex not in tent]
```
manasaV3, line 59: This check should be a part of the fixture that generates the run_names and not be done here. [Which it already is](https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/1deb4ab6c71c3347ee9b77c5fdcff4b412ce62c8/ingestion_tools/scripts/data_validation/conftest.py#L32-L34).

ingestion_tools/scripts/data_validation/conftest.py
```diff
@@ -35,35 +36,56 @@ def run_names(bucket: str, dataset: str, run_glob: str) -> List[str]:
...
     fs: S3Filesystem = FileSystemApi.get_fs_api(mode="s3", force_overwrite=False)
-    tentatives = fs.glob(f"s3://{bucket}/{dataset}/*/Tomograms/{voxelspacing_glob}")
+
+    tentatives = []
+
+    # Use ThreadPoolExecutor to run glob_task in parallel
+    with concurrent.futures.ThreadPoolExecutor() as executor:
+        # Map the run_names to the glob_task function
+        results = executor.map(
+            lambda run_name: fs.glob(f"s3://{bucket}/{dataset}/{run_name}/Tomograms/{voxelspacing_glob}"),
```
manasaV3, line 48: nit: could we create a formatted string outside the lambda, which already has the bucket, dataset, voxelspacing_glob applied? So, that this formatting only has to apply the run name?

ingestion_tools/scripts/data_validation/tests/helper_mrc_zarr.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def test_zarr_mrc_volume_size(self):
+        """Check that the zarray shape (volume size) matches the mrc volume size, for each zarr binning."""
+
+        def check_volume_size(header_data, zarr_filename):
+            mrc_file = zarr_filename.replace(".zarr", ".mrc")
+
+            zarrays = header_data["zarrays"]
+            for i, zarray in zarrays.items():
+                header = self.mrc_headers[mrc_file].header
```
manasaV3, line 28: nit: Could this header be outside the for loop?

ingestion_tools/scripts/data_validation/tests/test_tiltseries.py
```diff
@@ -0,0 +1,122 @@
...
+        assert len(set(tiltseries_mdoc["PixelSpacing"])) == 1
+        assert tiltseries_metadata["pixel_spacing"] == tiltseries_mdoc["PixelSpacing"][0]
+
+    def test_tiltseries_image_size_mdoc(self, tiltseries_metadata: Dict, tiltseries_mdoc: pd.DataFrame):
+        """Check that the tiltseries image size matches the MDOC data."""
+        assert len(set(tiltseries_mdoc["ImageSize"])) == 1
+        assert tiltseries_metadata["size"]["x"] == tiltseries_mdoc["ImageSize"][0][0]
+        assert tiltseries_metadata["size"]["y"] == tiltseries_mdoc["ImageSize"][0][1]
+
+    ### END MDOC consistency tests ###
```
manasaV3, line 122: While the mdoc files are saved along with the tiltseries at the moment, they are associated to the frames. As, there is processing to go from frames to tiltseries, it is very unlikely for us to have the same Image size or pixel spacings. So, these two tests are more likely to fail and that is the expected behaviour.

manasaV3 (changes requested):

ingestion_tools/scripts/data_validation/conftest.py
daniel-ji: fixed in a later pr!

ingestion_tools/scripts/data_validation/conftest.py
daniel-ji: fixed in a later pr, see #250

ingestion_tools/scripts/data_validation/conftest.py
daniel-ji: done in #250!

ingestion_tools/scripts/data_validation/tests/helper_mrc_zarr.py
daniel-ji: fixed!

ingestion_tools/scripts/data_validation/tests/test_tiltseries.py
daniel-ji: Would it be of interest to see which runs fail against these tests though? Maybe worth asking Utz tomorrow?

manasaV3 (approved):

PR #225, opened 2024-08-27
Adding sync to deposition

manasaV3:
Relates to 

## Description
Adds sync capability for deposition folders to the enqueue script. 

Example:
```
python enqueue_runs.py sync --include-deposition 10301 --dryrun --no-sync-dataset
```
The command above should do a dryrun to sync only the deposition metadata for deposition_id = 10301


jgadling (approved):
Eventually I think we should also support finding/syncing *all* depositions and / or filtering a list of all possible depositions to sync, the same way we support finding / filtering datasets. This is a good start though!

PR #227, opened 2024-08-28
feat: s3 data validation: tomograms & voxel spacing

daniel-ji:
To be merged after #229

See the voxel spacing and tomograms section of https://docs.google.com/document/d/1yMKM0DW9KRhlcYiBGPcR7oW0liGtUew6NAmBhMg5U3w/edit#heading=h.l8vgulbjis4

PR #230, opened 2024-08-28
feat: s3 data validation: cli improvements

daniel-ji:
To be merged after #227. Based heavily off of @uermel's previous work. 

Improves the s3 data validation command line interface / experience.
- Adds pytest-xdist plugin
- Adds documentation on how to run s3 data validation
- Small fix: voxelspacing_glob -> voxel_spacing_glob
- Cleans up pytest marks (metadata is not needed, is redundant / should be implied). The actual line of what is metadata validation can become a bit fuzzy sometimes when cross-checks with metadata & actual data occur. If self-contained, only metadata validation is desired, we can discuss this and build out support for it.


manasaV3 (approved):
lgtm

PR #213, opened 2024-08-21
Schema features & fixes: annotation object id UniProtKB support, annotation publications, microscope manufacturer enum

daniel-ji:
- Fix annotation publications list description (include PDB) and also have "PDB-" prefix, not "pdb" prefix
- Add "SIMULATED" option to microscope manufacturer enum
- Add UniProtKB id support to dataset.annotation_object.id
- Change `default_range` of v2.0.0 common.yaml schema to Any, so that `range` / type values of slots are properly written.

PR #231, opened 2024-08-28
Rename db comment for annotation object_id 

daniel-ji:
Follows up on supporting UniProtKB (which was done here https://github.com/chanzuckerberg/cryoet-data-portal-backend/pull/213)

Question: how do I actually get the migration run?


manasaV3:
I can work with you on releasing it to prod but merging it should run the migration in staging.

manasaV3 (approved):

PR #234, opened 2024-08-29
feat: support hff mesh ingestion

Bento007:
Relates to https://github.com/chanzuckerberg/cryoet-data-portal/issues/951
Blocked by #209 

## Description
- add support for ingest the hff mesh ingest format.
- add mesh_name to identify what mesh to extract from the hff file. This is because the hff format can store more than one mesh
- add AbstractTriangularMeshAnnotation to contain shared members across classes.
- Add TriangularMeshAnnotationGroup to ingest a mesh annotation where the mesh is stored among multiple meshes. 
- Add script to generate test hff data from a .glb mesh file. This is for reproducibility in the future if new test data needs to be generated.
- Added unit test for TriangularMeshAnnotationGroup.
- Update the ingestion_config schema with 

## Testing
- added unit tests
- tested with a real hff file and verified the generated hff had the same types and general structure.


ingestion_tools/scripts/common/mesh_converter.py
```diff
@@ -15,21 +17,91 @@ def convert_mesh_to_glb(convert_func: typing.Callable[[str], trimesh.Trimesh]) -
...
+        for mesh_list in fp["segment_list/"].keys(): # noqa
+            # Mesh name
+            mesh_name_raw: bytes = fp[f"segment_list/{mesh_list}/biological_annotation/name"][()]
+            if not mesh_name_raw:
+                logger.warning("No mesh name found at segment_list/{mesh_list} in the hff file")
+            else:
+                mesh_name: str = mesh_name_raw.decode()
+                if mesh_name == name:
+                    return mesh_list
+        raise ValueError(f"Mesh name {name} not found in the hff file")
```
manasaV3, line 46: There not being a mesh for a specific name, is a valid case. This should be caught by the `is_valid` check and not processed. So, we shouldn't throw and error here. But, it would be good to have a warning log.

ingestion_tools/scripts/common/mesh_converter.py
```diff
@@ -15,21 +17,91 @@ def convert_mesh_to_glb(convert_func: typing.Callable[[str], trimesh.Trimesh]) -
...
+
+            # Do the same for the vertices
+            vertices_data_path = f"segment_list/{mesh_index}/mesh_list/{mesh}/vertices"
+            vertices = _extract(fp, vertices_data_path)
+
+            # align the vertices and triangles
+
+
+            # Create a trimesh object from the vertices and triangles data
+            mesh = trimesh.Trimesh(vertices=vertices, faces=triangles, face_colors=color)
```
manasaV3, line 97: As we don't add color to the other trimeshes we create, this might be inconsistent. Also, we might not retain the same color in the neuroglancer visualization, that could create additional confusion.

ingestion_tools/scripts/common/mesh_converter.py
```diff
@@ -15,21 +17,91 @@ def convert_mesh_to_glb(convert_func: typing.Callable[[str], trimesh.Trimesh]) -
...
-        mesh: trimesh.Trimesh = convert_func(input_file)
+        mesh: trimesh.Trimesh = convert_func(input_file, *args, **kwargs)
         mesh.apply_scale(scale_factor)
         mesh.export(output_file)
 
     return wrapper
 
 
+def check_mesh_name(input_file: str, name: str) -> str:
+    with h5py.File(input_file, "r") as fp:
```
manasaV3, line 36: Could we add a comment that links to the schema documentation for h5?

ingestion_tools/scripts/tests/data_generators/generate_mesh.py
```diff
@@ -0,0 +1,60 @@
...
+"""
+
+import base64
+
+import h5py
+import numpy as np
+import trimesh
+
+
+def generate_hff_mesh(input_file: str, output_file: str) -> None:
```
manasaV3, line 12: 🔥

ingestion_tools/scripts/tests/data_generators/generate_mesh.py
```diff
@@ -0,0 +1,60 @@
...
+    cd = os.path.dirname(__file__).split("/")[0:-1]
+    test_data_location = os.path.join("/", *cd, "fixtures/annotations")
+
+    # create the input and output file paths
+    file_prefix = os.path.join(test_data_location, "triangular_mesh")
+    input_file = f"{file_prefix}.glb"
+    output_file = f"{file_prefix}.hff"
+
+    # generate the hff mesh file
+    generate_hff_mesh(input_file, output_file)
```
manasaV3, line 60: Did we want to retain this?

Bento007: If you're talking about the script, yes I think we should retain it. It could be helpful in the future for generating new hff files and for knowing debugging issues with hff conversion.

ingestion_tools/scripts/common/mesh_converter.py
```diff
@@ -15,21 +17,91 @@ def convert_mesh_to_glb(convert_func: typing.Callable[[str], trimesh.Trimesh]) -
...
+        for mesh_list in fp["segment_list/"].keys(): # noqa
```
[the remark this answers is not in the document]
Bento007: There shouldn't be but it's a possibility. I'll handle that case.

ingestion_tools/scripts/common/mesh_converter.py
Bento007: That information is in the from_hff function.

ingestion_tools/scripts/common/mesh_converter.py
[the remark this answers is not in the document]
Bento007: If this does happen a key error will be thrown which names the missing key name. How does that sounds?

ingestion_tools/scripts/common/mesh_converter.py
Bento007: should is_valid return true or false if it is not found?

schema/core/v1.1.0/metadata.yaml
```diff
@@ -641,6 +641,18 @@ classes:
       scale_factor:
         exact_mappings:
           - cdp-common:annotation_source_file_scale_factor
+  AnnotationTriangularMeshGroupFile:
+    is_a: AnnotationSourceFile
+    aliases:
+      - TriangularMeshGroup
+    description: File and sourcing data for a triangular mesh annotation. Annotation that identifies an object.
```
Bento007, line 648:
```suggestion
    description: File and sourcing data containing one or more triangular mesh annotations.
```

ingestion_tools/scripts/common/mesh_converter.py
Bento007: removed

ingestion_tools/scripts/common/mesh_converter.py
manasaV3: Yes, that is correct.

ingestion_tools/scripts/tests/data_generators/generate_mesh.py
Bento007: I'll keep it since it's a simple runnable example on how to generate the hff

PR #244, opened 2024-08-30
chore: Updating configs and validation

manasaV3:
## Description
- Updates the method_links field name to `custom_name` to prevent confusions
- Updates the common v2 schema to match field names with v2 api schema
- Adds the method_links to the api_schema


daniel-ji (approved):
looks good to me!

PR #237, opened 2024-08-30
feat: s3 data validation: setup allure and bugfixes for previous s3 data validation PR bugs

daniel-ji:
## setup allure
- add script to run allure tests
- add allure titles to everything (and make test names consistent)
- add documentation on how to run allure tests

## small fixes
- various, small, non-breaking fixes / renames
- clean up type hints

## actually modified tests in this PR (not just some name refactors, like many of the modified lines in this PR):
- restrain segmentation mask ZARR to only np.int8 (utz fixed all of them to be int8 only, no more float32)
	- Thank you Utz!
	- note mrc is still not yet fixed yet
- Delete a voxel spacing test (that checks to make sure our voxel spacing folders are named correctly) that is now covered by the fact that voxel spacings are stored as strings, so we will preserve the voxel spacing folder and ensure it is 3 decimal digits

PR #248, opened 2024-08-31
feat: s3 data validation: neuroglancer checks, other small validation tests

daniel-ji:
To be merged in after #237. 

- add neuroglancer validation (simple checks), see https://docs.google.com/document/d/1yMKM0DW9KRhlcYiBGPcR7oW0liGtUew6NAmBhMg5U3w/edit#heading=h.sb58fesxyei5
- add run deposition id validation (only if the deposition id exists, currently allowing it to not exist and the test just gets skipped)
- check mrc volume size is consistent with header data


ingestion_tools/scripts/data_validation/fixtures/path.py
```diff
@@ -305,12 +305,28 @@ def tomo_zarr_file(
...
+# =============================================================================
+
+
 @pytest.fixture(scope="session")
-def tomo_basename(
-    tomo_zarr_file: str,
+def neuroglancer_dir(tomo_dir: str, filesystem: FileSystemApi) -> str:
+    """[Dataset]/[ExperimentRun]/Tomograms/VoxelSpacing[voxel_spacing]/CanonicalTomogram/"""
+    return tomo_dir
+
```
manasaV3, line 317: This is very nice, as it gives us opportunity to easily make any updates in the future.

ingestion_tools/scripts/data_validation/tests/test_neuroglancer.py
```diff
@@ -0,0 +1,19 @@
...
+
+from common.fs import FileSystemApi
+
+
+@pytest.mark.tomogram
+@pytest.mark.parametrize("run_name, voxel_spacing", pytest.run_spacing_combinations, scope="session")
+class TestNeuroglancer:
+    @allure.title("Neuroglancer: sanity check neuroglancer config file.")
+    def test_metadata(self, neuroglancer_config: Dict):
+        assert len(neuroglancer_config) > 0
```
manasaV3, line 15: Can we update this test to test there is at least one layer in the config? 
```suggestion
        assert "layers" in  neuroglancer_config and  len(neuroglancer_config["layers"]) > 0
```

ingestion_tools/scripts/data_validation/tests/test_neuroglancer.py
```diff
@@ -0,0 +1,19 @@
...
+
+    @allure.title("Neuroglancer: valid corresponding deposition metadata.")
+    def test_deposition_id(self, neuroglancer_config: Dict, bucket: str, filesystem: FileSystemApi):
+        HelperTestDeposition.check_deposition_metadata(neuroglancer_config["deposition_id"], bucket, filesystem)
```
manasaV3, line 19: We can remove this check, as it is not required for the config to have the deposition_id.

manasaV3 (changes requested):

manasaV3 (approved):
🚢

PR #250, opened 2024-09-02
feat: s3 data validation: Support running pytest across multiple datasets

daniel-ji:
Previously, we would have to call pytest multiple times for multiple datasets. Now, one pytest call can take in multiple datasets (via `--datasets`) and tests will run on all the specified datasets in one go.

Refactors conftests.py and cleans up the code in addition to adding this new feature.

And also a bunch of bug fixes 😭


ingestion_tools/scripts/data_validation/run_tests.py
```diff
@@ -10,79 +9,95 @@
...
 
         # Generate the report
         os.system(f"allure generate --output {localdir_rep} {localdir_raw}")
 
         # Upload the new report
-        if save_history:
-            os.system(f"tar -czf {local_tar} -C {localdir_rep} .")
-            fs.s3fs.put(local_tar, remote_dataset_tar)
-            os.system(f"rm -rf {local_tar}")
+        fs.s3fs.put(localdir_rep, remote_dataset_dir, recursive=True)
```
manasaV3, line 100: Why not use `fs.push`?

ingestion_tools/scripts/data_validation/run_tests.py
```diff
@@ -10,79 +9,95 @@
...
-        remote_dataset_dir = f"{output_bucket}/data_validation/{dataset}"
-        remote_dataset_tar = f"{remote_dataset_dir}/{dataset}.tar.gz"
-        fs.makedirs(remote_dataset_dir)
-
-        if fs.exists(remote_dataset_tar):
-            get_history(remote_dataset_tar, localdir_raw, fs)
+        remote_dataset_dir = f"{output_bucket}/{output_dir}/{dataset}"
+        if fs.exists(remote_dataset_dir):
+            get_history(remote_dataset_dir, localdir_raw, fs)
+            fs.s3fs.rm(remote_dataset_dir, recursive=True)
```
manasaV3, line 94: As there is a timestamp associated to it, what would be a usecase for this to already exist?

manasaV3 (changes requested):

ingestion_tools/scripts/data_validation/run_tests.py
daniel-ji: To keep track of history, we currently just upload the latest report run (with the history/ folder in it) for a given dataset. We don't track every single report, but instead just the latest report (which holds the history of the last 20 reports before it). Do we want to keep every report that is run?

ingestion_tools/scripts/data_validation/run_tests.py
daniel-ji: fs.push requires it to be in the temp dir, but the user may want to store it somewhere other than the tmp dir?

PR #253, opened 2024-09-04
chore: Updating deposition types

manasaV3:
## Description
Updates deposition types to include annotation for dataset depositions that include annotation in them.


kne42 (approved):
lmk when changes are available in prod so I can replace our queries

uermel (approved):
LGTM!

PR #257, opened 2024-09-05
feat: Update ingestion configs with collection_metadata entities

Bento007:
relates to #997

## Description

- Add collection_metadata to the ingest config importer, schema and template
- Remove mdoc files from rawtilts and move to collection_metadata


ingestion_tools/scripts/standardize_dirs.py
```diff
@@ -58,6 +59,7 @@
                 FrameImporter: {},
                 TiltSeriesImporter: {},
                 RawTiltImporter: {},
+                CollectionMetadataImporter: {},
```
manasaV3, line 62: This also needs to be added to https://github.com/chanzuckerberg/cryoet-data-portal-backend/pull/257/files#diff-a112fce9544e4d2a0ebafa1a467c88c2ba702ec321b75651e92683d9fd319d6bL24 for it to support the relevant import flags.

ingestion_tools/scripts/importers/collection_metadata.py
```diff
@@ -0,0 +1,8 @@
+from common.finders import DefaultImporterFactory
+from importers.base_importer import BaseFileImporter
+
+
+class CollectionMetadataImporter(BaseFileImporter):
```
manasaV3, line 5: We should add an entry to https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/99adb45f4c7a42a52da9f7949ee7fc6caedeba03/ingestion_tools/scripts/common/config.py#L243
for `collection_metadata`, with the same value as `rawtilts`.

ingestion_tools/scripts/importers/collection_metadata.py
```diff
@@ -0,0 +1,8 @@
...
+    type_key = "collection_metadata"
```
manasaV3, line 6: This also needs a plural key.
```suggestion
    type_key = "collection_metadata"
    plural_key = "collection_metadata"
```

manasaV3 (changes requested):

ingestion_tools/dataset_configs/10000.yaml
```diff
@@ -333,6 +333,11 @@ annotations:
       glob_string: labels/{run_name}_membranes.mrc
       is_visualization_default: false
       mask_label: 1
+collection_metadata:
+- sources:
+  - source_multi_glob:
+      list_globs:
+      - metadata/mdocs_modified/{run_name}.mdoc
```
manasaV3, line 340: nit: This could be `source_glob` as we only have one entry.

ingestion_tools/scripts/standardize_dirs.py
Bento007: can you elaborate. I'm not sure what you mean by this?

ingestion_tools/scripts/common/config.py
```diff
@@ -266,6 +266,8 @@ def resolve_output_path(self, key: str, obj: BaseImporter) -> str:
             "deposition_metadata": "depositions_metadata/{deposition_name}/deposition_metadata.json",
             "deposition_keyphoto": "depositions_metadata/{deposition_name}/Images",
         }
+        paths["collection_metadata"] = paths["rawtilt"]
```
manasaV3, line 269: Could we hardcode this in the map for consistency?

ingestion_tools/scripts/standardize_dirs.py
manasaV3: For us to have an `--import-collection-metadata` flag that is supported by ingestion process, we need to have this importer class listed in the `IMPORTERS` constant.

schema/ingestion_config/v1.0.0/ingestion_config_models.yaml
```diff
@@ -54,6 +54,10 @@ classes:
         description: Annotations for the dataset.
         multivalued: true
         range: AnnotationEntity
+      collection_metadata:
```
daniel-ji, line 57: Extended validation for this should be added as well - I can hop on a call and show you how to do this, if you'd like. Thanks!

ingestion_tools/scripts/common/config.py
Bento007: I think the paths variable should live closer the classes that use them. Updating this file when adding a new ingest config entity is not intuitive at all.

manasaV3 (approved):
🚢

PR #259, opened 2024-09-06
add extended validation check for collection metadata

daniel-ji:
The rawtilts entry for 10010.yaml was also failing (the list_globs array must have at least one element if it exists), so i just commented it out. We can also remove it, if that seems fit.


Bento007 (approved):
Awesome, thank you for writing this.

PR #263, opened 2024-09-10
feat: S3 Ingestion workflow for Alignment 

manasaV3:
Relates to https://github.com/chanzuckerberg/cryoet-data-portal/issues/898

## Description

- Refactors the AnnotationIdentifierHelper to be more generic
- Adds the default config value conditionally such that it only gets added if there isn't a pre-existing entry for it in the config
- Reorders the processing of Gains, Frames, CollectionMetadata, RawTilt, TiltSeries, Alignment to be before Tomogram
- Creates a new alignment importer that also handles metadata generation for alignment. 
  - The alignment files are copied over with an id prefix to the /Alignment folder at the Run level
  - The alignment files are only processed if either they have volume dimensions in the metadata or have an associated tomogram from which they volume dimension can be sourced
- Replaces the fake tomogram files with ones valid tomgram files, and a script to create them.

#### per_section_alignment_parameters
- This will be generated from the source alignment, tilt and xtilt files. This code for that is encapsulated such that @uermel  can make updates the logic without requiring changes to the metadata generation process.
- This will be empty for default alignment
- The values for any data whose source is not available be 0

#### Sample alignment metadata

```
{
    "affine_transformation_matrix": [
        [
            1,
            0,
            0,
            0
        ],
        [
            0,
            1,
            0,
            0
        ],
        [
            0,
            0,
            1,
            0
        ],
        [
            0,
            0,
            0,
            1
        ]
    ],
    "alignment_type": "GLOBAL",
    "format": "xf",
    "is_canonical": true,
    "tilt_offset": 0,
    "volume_offset": {
        "x": 0,
        "y": 0,
        "z": 0
    },
    "x_rotation_offset": 0,
    "per_section_alignment_parameters": [
        {
            "z_index": 0,
            "tilt_angle": -41.01,
            "volume_x_rotation": 0,
            "in_plane_rotation": [
                0.1131234,
                -0.9935809,
                0.9935809,
                0.1131234
            ],
            "x_offset": 1498.255,
            "y_offset": -191.916
        },
        ...
    ],
    "alignment_path": "cryoetportal-output-test/10301/01122021_BrnoKrios_arctis_lam3_pos27/Alignments/100-01122021_BrnoKrios_arctis_lam3_pos27_dose-filt.xf",
    "tilt_path": "cryoetportal-output-test/10301/01122021_BrnoKrios_arctis_lam3_pos27/Alignments/100-01122021_BrnoKrios_arctis_lam3_pos27_dose-filt.tlt",
    "tiltx_path": null,
    "volume_dimension": {
        "x": 1024,
        "y": 1024,
        "z": 512
    },
    "deposition_id": 10300,
    "last_updated_at": 1726031170
}
```


ingestion_tools/scripts/common/config.py
```diff
@@ -16,28 +16,6 @@
     BaseImporter = "BaseImporter"
 
 
-APPEND_STATIC_CONFIG: str = """
```
Bento007, line 19: is this related to the PR?

manasaV3: Yeah, I cleaned up to make this more intutive as I was updating the default config value to be conditional.

ingestion_tools/scripts/tests/fixtures/dataset1.yaml
```diff
@@ -23,6 +23,21 @@ annotations:
     file_format: csv
     glob_string: particle_lists/{run_name}_fas.csv
     shape: Point
+alignments:
+- metadata:
+    affine_transformation_matrix: [[2, 0, 0, 0], [0, 3, 0, 0], [0, 4, 1, 0], [0, 0, 0, 5]]
```
Bento007, line 28: I don't think yaml understands this syntax.

It should be something like this
```YAML
affine_transformation_matrix: 
  - - 2
    - 0
    - 0
    - 0
   - - 0
     - 3
     - 0
     - 0
```

Bento007 (changes requested):

ingestion_tools/scripts/common/config.py
```diff
@@ -248,8 +228,10 @@ def resolve_output_path(self, key: str, obj: BaseImporter) -> str:
             "tiltseries": "{dataset_name}/{run_name}/TiltSeries",
             "gain": "{dataset_name}/{run_name}/Frames/{run_name}_gain",
```
Bento007, line 229: > ### Gains
> Update gain files to be stored in /<dataset-id>/<run-name>/Gain
> Update the naming of the gain file to be <name-in-source>.<ext>

ingestion_tools/scripts/common/config.py
```diff
@@ -248,8 +228,10 @@ def resolve_output_path(self, key: str, obj: BaseImporter) -> str:
...
             "frame": "{dataset_name}/{run_name}/Frames",
-            "rawtilt": "{dataset_name}/{run_name}/TiltSeries",
-            "collection_metadata": "{dataset_name}/{run_name}/TiltSeries",
+            "rawtilt": "{dataset_name}/{run_name}/Frames",
+            "collection_metadata": "{dataset_name}/{run_name}/Frames",
```
Bento007, line 232: should this have been part of [feat: Update ingestion configs with collection_metadata entities](https://github.com/chanzuckerberg/cryoet-data-portal-backend/pull/257)?

ingestion_tools/scripts/importers/tilt.py
```diff
@@ -0,0 +1,37 @@
...
+from common.finders import DefaultImporterFactory
+from importers.base_importer import BaseFileImporter
+
+if TYPE_CHECKING:
+    AlignmentIdentifierHelper = "AlignmentIdentifierHelper"
+else:
+    from importers.alignment import AlignmentIdentifierHelper
+
+
+class TiltImporter(BaseFileImporter):
```
Bento007, line 14: What is this class needed for?

Bento007 (commented):
Can you some tests that verify the files are stored in the expected location for `./Frames` and `./Gains`

ingestion_tools/scripts/importers/alignment.py
```diff
@@ -0,0 +1,180 @@
...
+    ) -> list:
+        result = []
+        if self.is_default_alignment():
+            return result
+        xf_data = self.get_xf_data()
+        tlt_data = self.get_dataframe(tlt_importer, ["tilt_angle"])
+        tltx_data = self.get_dataframe(tltx_importer, ["volume_x_rotation"])
+        for index, entry in xf_data.iterrows():
+            item = {
+                "z_index": index,
```
Bento007, line 117: Are you missing `beam_tilt`?

Bento007: > Are you missing `beam_tilt`?

NVM, I see NVNM I see https://github.com/chanzuckerberg/cryoet-data-portal/issues/898#issuecomment-2329765964

ingestion_tools/scripts/importers/alignment.py
```diff
@@ -0,0 +1,180 @@
...
+
+import pandas as pd
+
+from common.config import DepositionImportConfig
+from common.finders import DefaultImporterFactory
+from common.id_helper import IdentifierHelper
+from common.metadata import AlignmentMetadata
+from importers.base_importer import BaseFileImporter, BaseImporter
+
+if TYPE_CHECKING:
```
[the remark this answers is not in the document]
manasaV3: This isn't needed right now, but this is to future proof the circular dependency between Tomogram and Alignments.

manasaV3:
> Can you some tests that verify the files are stored in the expected location for ./Frames and ./Gains
The Frames are already being stored in the expected location.

The Gains' new location needed some confirmations, before being updated.

ingestion_tools/scripts/importers/tilt.py
manasaV3: No, this doesn't need to be as this was removed.

ingestion_tools/scripts/common/alignment_converter.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def alignment_converter_factory(
+    config: DepositionImportConfig,
+    metadata: dict[str, Any],
+    path: str,
+    parents: dict[str, Any],
+) -> BaseAlignmentConverter:
+    alignment_format = metadata.get("format")
+    if alignment_format == "xf":
```
jgadling, line 113: I thought that the idea for alignments was that in some cases we'd have *multiple* files to read to generate a portal alignment: some combination of [`tilt.com`, `*.xf`], or `*.aln`

manasaV3: I believe the transformation of multiple alignments into a single standardized format, was gonna be handled in the next phase.

jgadling: Do you think we can rearrange the ingestion configs for alignment so that there's room to add more source file types when we're ready, even if we only support one file per alignment for now?

ingestion_tools/scripts/common/id_helper.py
```diff
@@ -0,0 +1,79 @@
...
+    @classmethod
+    def get_identifier(
+        cls,
+        config: DepositionImportConfig,
+        metadata: dict[str, Any],
+        parents: dict[str, Any],
+        *args,
+        **kwargs,
+    ):
+        container_key = cls._get_container_key(config, parents, *args, **kwargs)
```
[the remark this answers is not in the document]
jgadling: Should we enable stronger linting?

ingestion_tools/scripts/importers/alignment.py
```diff
@@ -0,0 +1,180 @@
...
+        path: str,
+        parents: dict[str, Any],
+    ):
+        super().__init__(config, metadata, name, path, parents)
+        self.identifier = AlignmentIdentifierHelper.get_identifier(config, metadata, parents)
+
+    def get_dest_filename(self) -> str:
+        if not self.path:
+            return None
+        output_dir = self.get_output_path()
```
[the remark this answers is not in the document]
manasaV3: Unfortunately, we can't do this as the parent class does a join as a file path, while we just append the strings here.
I will move this method closer to the overridden definition of `get_output_path()` in the class to make this clearer.

ingestion_tools/scripts/common/config.py
manasaV3: While, this could have been, the updates to the existing paths were covered as a part of this issue.

ingestion_tools/scripts/common/alignment_converter.py
manasaV3: The intention is to use format to specify the types of files. Could you elaborate on the recommendation?

ingestion_tools/scripts/common/id_helper.py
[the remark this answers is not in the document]
Bento007: I find the enforced typing to be a bit over zealous. A bit more documentation of the functions, and the types would be enough.

Bento007 (commented):
Thank you for the docstring, those are very helpful.

ingestion_tools/scripts/common/config.py
```diff
@@ -250,6 +230,7 @@ def resolve_output_path(self, key: str, obj: BaseImporter) -> str:
             "frame": "{dataset_name}/{run_name}/Frames",
             "rawtilt": "{dataset_name}/{run_name}/TiltSeries",
             "collection_metadata": "{dataset_name}/{run_name}/TiltSeries",
+            "alignment": "{dataset_name}/{run_name}/Alignments",
```
Bento007, line 233: "gain" on line 249 needs to be updated as part of this work.

ingestion_tools/scripts/importers/alignment.py
```diff
@@ -0,0 +1,145 @@
...
+        volume_dim = self.metadata.get("volume_dimension", {})
+        return all(volume_dim.get(dim) for dim in "xyz") or next(
+            TomogramImporter.finder(self.config, **self.parents),
+            None,
+        )
+
+    @classmethod
+    def get_default_config(cls) -> list[dict] | None:
+        return [
+            {
```
Bento007, line 129: should these defaults align with the defaults [here](https://github.com/chanzuckerberg/cryoet-data-portal/issues/997)?

ingestion_tools/scripts/common/finders.py
```diff
@@ -218,3 +234,22 @@ def load(
         if source.get("literal"):
             return BaseLiteralValueFinder(**source["literal"])
         raise Exception("Invalid source type")
+
+
+class MultiSourceFileGlobFinder(DefaultImporterFactory):
```
jgadling, line 239: Nice! My only real comment is that this config is a little bit fragile - it lets us list multiple globs that can find different kinds of files to import for a single annotation. For example we can have config like this:
```
alignments:
- metadata:
    ....
  sources:
    source_multi_glob:
       list_globs:
         - "{run_name}/tilts/*.xf"
         - "{run_name}/tilts/tilt.com"
```
, but it doesn't let us specify anything EXTRA each glob. For example we can't specify a regex filter or format information for each glob. This means that we're 100% stuck relying on filename extensions to understand how to read each file returned by this finder. I think that most tomogram reconstruction workflows are pretty standardized and already heavily reliant on file names being named in special ways, but wanted to get a comment from @uermel on whether this is reasonable.

uermel: I can't guarantee that this will be sufficient in every case, pretty much every expectation I had concerning users' data organization has been proven wrong by an edge case. 

That said, I think within a group of runs globs will suffice. Will there be the ability to define include/exclude patterns for groups of runs? In that case one could define multiple alignments blocks for different groups of runs, where the glob patterns can change. I think that should be sufficient.

manasaV3: > Will there be the ability to define include/exclude patterns for groups of runs for the alignment blocks? In that case one could define multiple alignments blocks for different groups of runs, where the glob patterns can change. I think that should be sufficient. In the worst case we may have to use csv-mapped values to include in these glob strings.

Yes, we should be able to use parent filters like we do in other places. But, list globs don't support match_regex at the moment.

manasaV3: If we don't have any outliers for this at the moment, as we comfortable with this solution for now, and maybe extending it in the future when the need arises? @jgadling

jgadling: Ok, now that `AlignmentImporter` has a custom `ImporterFactory` I have two concerns about where it's at right now:
1. It takes exactly the same `sources` configs as our default `ImporterFactory` but it has different *behavior* - I think this is confusing -- for tiltseries for example, every file found by every glob in the `source_multi_glob` finder will become a new Tiltseries, but for Alignments, the files found by every glob in the `source_multi_glob` config will all be *combined* into a single Alignment. I'd prefer if the interface for the alignment configs somehow let users know that it works *differently* for alignments than it does for other imported types.
2. If (when) we find outliers in the future, my concern is that we'll have to change the format of the `sources` blocks in all existing ingestion configs to match the format supported by our newer/better/outlier-handling finder. It's not a lot of work to write a finder with a more flexible config structure right now, but it's a lot more work migrating all this later.
Maybe something along the lines of:
```yaml
  sources:
  - fan_in_finder:
    - glob: "{run_name}/tilts/*.xf"
      include_regex: ".*FILTERED.*"
      format: "xf"
    - glob: "{run_name}/tilts/tilt.com"
      format: "tilt.com"
```
**Caveat though** -- Now that `Alignments` has a custom finder, we have the ability to add **additional** source types without breaking reverse compatibility, so item 2 isn't a hard blocker, but I'm curious to know what you think.

manasaV3: As the new importer is handling the usecases for current and foreseeable future, I think this might be good enough to get us to the finish line. Especially as it only has internal users who would be interacting with it at the moment who are aware of the alignment ingestion behaviour.  

I was also originally envisioning something with a key that defines the  explicitly for the finder. But, that doesn't provide a lot of added benefit right now, so I would lean towards working on this enhancement in the next phase. 

I believe there shouldn't be a big enough difference in the amount of effort needed to migrate the work later to newer pattern between doing it now or later, as it is being done in two steps, instead of one single step.

ingestion_tools/scripts/common/config.py
manasaV3: Addressed in https://github.com/chanzuckerberg/cryoet-data-portal-backend/pull/271

test_infra/test_files/input_bucket/10001_input/tomograms/TS_run1.rec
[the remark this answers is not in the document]
manasaV3: Added https://github.com/chanzuckerberg/cryoet-data-portal-backend/pull/263/files#diff-66c071b6f5d9921c91d98c3fa516d013c5aceb78a8574f5a1a2ef3da468f1df9R1 to capture this.

ingestion_tools/scripts/common/config.py
manasaV3: Addressed in https://github.com/chanzuckerberg/cryoet-data-portal-backend/pull/271

ingestion_tools/scripts/common/finders.py
jgadling: I'd prefer if we handled these cases now, but if you want to proceed with this as it is, I think we need to *document* the differences between the multi-glob finder behavior in Alignments vs other importers. Maybe a big code block in `alignments.py` or maybe a markdown file. The "behavior is weird but it's just us" thing falls apart as we onboard more/different people to the project, so we need to leave some notes behind.

jgadling (approved):
Approving this to unblock a merge, but please add docs before merging!

uermel (approved):
Looks good from my side as well.

PR #270, opened 2024-09-13
feat: support importing s3 files into the v2 database

jgadling:
This is basically a straight port of the code referenced by `ingestion_tools/scripts/db_import.py` to use the schema and models defined by the apiv2 application.

Most of the differences between this code and the v1 code are:
1. Everything's been updated to use SQLAlchemy instead of peewee, and we're using the codegen'd models from apvi2
2. The test config uses the db test config set up by apiv2, which uses [pytest-postgresql](https://pypi.org/project/pytest-postgresql/) to create a temporary schema in a running postgres instance for each test. This means that we can run tests in parallel!

I also added some extra flags to the schema so that we use SQLAlchemy's default [cascade delete](https://docs.sqlalchemy.org/en/20/orm/cascades.html) behavior to delete child rows when a parent row is deleted.


.gitattributes
```diff
@@ -0,0 +1,7 @@
+apiv2/cerbos/** linguist-generated
+apiv2/database/** linguist-generated
+apiv2/graphql_api/** linguist-generated
+apiv2/support/** linguist-generated
+apiv2/test_infra/** linguist-generated
+apiv2/validators/** linguist-generated
+schema/**/*materialized*.yaml linguist-generated
```
jgadling, line 7: I'm hoping this hides some codegen'd files by default when looking at PR's

apiv2/conftest.py
```diff
@@ -64,57 +63,39 @@ def get_db_uri(
...
+@pytest.fixture()
+def test_db_uri(test_db: NoopExecutor) -> str:
+
+    return get_db_uri(
+        db_host=test_db.host,
+        db_port=test_db.port,
+        db_user=test_db.user,
+        db_pass=test_db.password,
+        db_name=test_db.dbname,
+    )
```
jgadling, line 78: Just some cleanup.

apiv2/db_import/importer.py
```diff
@@ -0,0 +1,225 @@
...
+    import_depositions: bool = False,
+    import_runs: bool = False,
+    import_tiltseries: bool = False,
+    import_tomograms: bool = False,
+    import_tomogram_authors: bool = False,
+    import_tomogram_voxel_spacing: bool = False,
+    import_everything: bool = False,
+    deposition_id: list[str] | None = None,
+    endpoint_url: str | None = None,
+):
```
jgadling, line 112: I made this a callable function so we don't have to invoke click in our tests. Things are clunky either way, but click was hiding tracebacks in our test results.

apiv2/schema/schema.yaml
```diff
@@ -553,13 +553,17 @@ classes:
         multivalued: true
         inverse: AnnotationFile.alignment
         range: AnnotationFile
+        annotations:
+          cascade_delete: true
```
jgadling, line 557: I tried to take a middle-ground approach in choosing which objects to delete when the parents got deleted, but this could use a look from @manasaV3 / @uermel

apiv2/template_overrides/support/enums.py.j2
```diff
@@ -0,0 +1,19 @@
+"""
+GraphQL enums
```
jgadling, line 2: I forgot to commit this file earlier, but the changes it makes (converting spaces to underscores in enum values) are already in the `main` branch.

test_infra/test_files/30001/RUN2/Tomograms/VoxelSpacing3.456/CanonicalTomogram/tomogram_metadata.json
```diff
@@ -1,6 +1,6 @@
 {
     "voxel_spacing": 3.456,
-    "reconstruction_method": "",
+    "reconstruction_method": "SART",
```
jgadling, line 3: Just making sure our metadata is valid.

apiv2/db_import/importers/base_importer.py
```diff
@@ -0,0 +1,299 @@
...
+        return all_records, hashed_records
+
+    def update_data_map(self, data_map: dict[str, Any], metadata: dict[str, Any], index: int) -> dict[str, Any]:
+        """
+        Method to be overridden in cases where custom data mapping is necessary based on index value of the record in
+        the list.
+        """
+        return data_map
+
+    def import_to_db(self) -> Base:
```
jgadling, line 204: Most of the code copied over from the v1 schema is just minor tweaks to reflect schema and sqlalchemy changes, but I made pretty significant changes to this `StaleDeletionDBImporter` class since it wasn't handling cases where there were multiple rows for a hash key. This could use 👀 and comparison to the old one.

PR #271, opened 2024-09-13
feat: Updating paths for Gains and CollectionMetadata

manasaV3:
Relates to https://github.com/chanzuckerberg/cryoet-data-portal/issues/898

### Frames
- Update the location of .mdoc to `/<dataset-id>/<run-name>/Frames`
### Gains
- Update gain files to be stored in `/<dataset-id>/<run-name>/Gain`
- Update the naming of the gain file to be `<name-in-source>.<ext>`


Bento007 (approved):

ingestion_tools/scripts/tests/fixtures/dataset1.yaml
```diff
@@ -23,6 +23,11 @@ annotations:
     file_format: csv
     glob_string: particle_lists/{run_name}_fas.csv
     shape: Point
+collection_metadata:
```
jgadling, line 26: is the code for collection_metadata support not part of this PR?

jgadling (approved):
Seems like there's an extra file in thIs PR?

ingestion_tools/scripts/tests/fixtures/dataset1.yaml
manasaV3: Good catch! the collection_metadata test file got missed in being included.

PR #273, opened 2024-09-17
feat: update core schema v2 for alignments and mesh

Bento007:
## Description
Update core schema v2 to match the changes in for v1 from https://github.com/chanzuckerberg/cryoet-data-portal/issues/997

- remove is_portal_standard.


jgadling (commented):
Please run `make-build-ingestion-config` to update the pydantic models before merging!

schema/core/v2.0.0/metadata.yaml
```diff
@@ -532,6 +532,24 @@ classes:
...
+        description: Whether the tomogram is a portal standard.
+        range: boolean
+        required: false
+        ifabsent: false
+      is_visualization_default:
+        description: Whether the tomogram is the default for visualization.
+        range: boolean
+        required: true
+        ifabsent: true
+      cross_references:
```
jgadling, line 545: Just calling this out because both old and new API's are missing field for tomogram cross references and dates. Is this PR intentionally adding functionality?

schema/core/v2.0.0/metadata.yaml
```diff
@@ -594,6 +612,9 @@ classes:
       is_visualization_default:
         exact_mappings:
           - cdp-common:annotation_source_file_is_visualization_default
+      is_portal_standard:
+        exact_mappings:
+          - cdp-common:annotation_source_file_is_portal_standard
```
jgadling, line 617: We had a discussion about the `is_portal_standard` field the other day and iirc It's only supposed to be present on `Tomogram` and `Alignment`

jgadling: cc @uermel

jgadling (changes requested):
Just marking this for changes while we get clarification on the crossreferences/dates fields on tomograms, and the `is_portal_standard` field on annotations. The rest lgtm!

schema/core/v2.0.0/metadata.yaml
Bento007: we do want these

schema/core/v2.0.0/metadata.yaml
Bento007: remove from v1 and v2

jgadling (approved):
Approved once the standardized field is removed from annotations

manasaV3 (approved):

PR #275, opened 2024-09-17
feat: support importing s3 files into the v2 database

jgadling:
This is basically a straight port of the code referenced by `ingestion_tools/scripts/db_import.py` to use the schema and models defined by the apiv2 application.

**TO REVIEW THIS PR** I recommend clicking this link: https://github.com/chanzuckerberg/cryoet-data-portal-backend/pull/275/commits/4b00b56084470a239b2a7ceb24f48a69311e635c
- The first commit in this PR is just copying the apiv1 files into the apiv2 directory, totally unchanged
- The second commit (linked above) is all of the changes made to support apiv2 db ingestion

Most of the differences between this code and the v1 code are:
1. Everything's been updated to use SQLAlchemy instead of peewee, and we're using the codegen'd models from apvi2
2. The test config uses the DB test config set up by apiv2, which uses [pytest-postgresql](https://pypi.org/project/pytest-postgresql/) to create a temporary schema in a running postgres instance for each test. This means that we can run tests in parallel!

I also added some extra flags to the schema so that we use SQLAlchemy's default [cascade delete](https://docs.sqlalchemy.org/en/20/orm/cascades.html) behavior to delete child rows when a parent row is deleted.


apiv2/support/enums.py
```diff
@@ -11,21 +11,21 @@
 
 
 @strawberry.enum
-class annotation_file_source_enum(enum.Enum):
+class annotation_file_source_enum(enum.StrEnum):
```
manasaV3, line 14: I believe we should also handle for case where an annotation submitted by the dataset_author is the portal_standard.

apiv2/support/enums.py
```diff
@@ -11,21 +11,21 @@
...
     dataset_author = "dataset_author"
     community = "community"
     portal_standard = "portal_standard"
 
 
 @strawberry.enum
-class annotation_method_type_enum(enum.Enum):
+class annotation_method_type_enum(enum.StrEnum):
```
manasaV3, line 21: nit: iirc believe we are adding an additional type here `simulated`.

apiv2/support/enums.py
```diff
@@ -62,35 +62,35 @@ class sample_type_enum(enum.Enum):
...
+class tiltseries_camera_acquire_mode_enum(enum.StrEnum):
     counting = "counting"
     superresolution = "superresolution"
     linear = "linear"
     cds = "cds"
 
 
 @strawberry.enum
-class tiltseries_microscope_manufacturer_enum(enum.Enum):
+class tiltseries_microscope_manufacturer_enum(enum.StrEnum):
```
manasaV3, line 73: Should this be an enum at the db level?

apiv2/support/enums.py
```diff
@@ -62,35 +62,35 @@ class sample_type_enum(enum.Enum):
...
-class tomogram_processing_enum(enum.Enum):
+class tomogram_processing_enum(enum.StrEnum):
     denoised = "denoised"
     filtered = "filtered"
     raw = "raw"
 
 
 @strawberry.enum
-class tomogram_reconstruction_method_enum(enum.Enum):
+class tomogram_reconstruction_method_enum(enum.StrEnum):
```
manasaV3, line 93: nit: Additional option "WBP", "Unknown"

test_infra/test_files/30001/RUN2/Tomograms/VoxelSpacing3.456/CanonicalTomogram/tomogram_metadata.json
```diff
@@ -1,6 +1,6 @@
 {
     "voxel_spacing": 3.456,
-    "reconstruction_method": "",
```
manasaV3, line 3: Is there any updates required for the v1 db_import related to these metadata updates?

apiv2/database/models/tomogram_voxel_spacing.py
```diff
@@ -50,6 +50,7 @@ class TomogramVoxelSpacing(Base):
         back_populates="tomogram_voxel_spacing",
         uselist=True,
         foreign_keys="Tomogram.tomogram_voxel_spacing_id",
+        cascade="all, delete-orphan",
```
manasaV3, line 53: Should this also delete the orphaned annotation files?

apiv2/db_import/importers/base_importer.py
```diff
@@ -0,0 +1,299 @@
...
+    """
+
+    def update_data_map(self, data_map: dict[str, Any], metadata: dict[str, Any], index: int) -> dict[str, Any]:
+        """Adds author list order based on the index position to the author data"""
+        if metadata.get("author_list_order"):
+            return data_map
+        return {**data_map, **{"author_list_order": index + 1}}
+
+
+class StaleParentDeletionDBImporter(StaleDeletionDBImporter):
```
manasaV3, line 250: I think this class might no longer be needed with the cascade delete in place.

manasaV3 (changes requested):
The cascade delete is cool.

apiv2/support/enums.py
jgadling: I'm trying to keep most of the schema changes (except what's specifically necessary to support the db import script) to the #277 PR - I'll add a note there!

apiv2/support/enums.py
jgadling: Also added this as a note on the other PR!

apiv2/support/enums.py
```diff
@@ -43,14 +43,14 @@ class annotation_method_link_type_enum(enum.Enum):
 
 
 @strawberry.enum
-class deposition_types_enum(enum.Enum):
+class deposition_types_enum(enum.StrEnum):
```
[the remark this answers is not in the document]
jgadling: We have a relationship between a Deposition and a DepositionType -- check `schema/schema.yaml` 
Basically a `Deposition` can have many `DepositionType` rows associated, and each `DepositionType.type` field has a value from this enum.

apiv2/support/enums.py
jgadling: Already in the other PR

apiv2/support/enums.py
jgadling: Yep! these enums are mapped to our SQLAlchemy models, like here for example:
https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/main/apiv2/database/models/tiltseries.py#L75-L77

apiv2/database/models/tomogram_voxel_spacing.py
jgadling: Ooh, good catch! updated!

test_infra/test_files/30001/RUN2/Tomograms/VoxelSpacing3.456/CanonicalTomogram/tomogram_metadata.json
jgadling: Nope, these are just helping the existing test files fit our schema better.

apiv2/db_import/importers/base_importer.py
jgadling: We still kind of do -- the interfaces are different for `StaleDeletionDBImporter` and `StaleParentDeletionDBImporter` - the first one expects to import all child rows at once, and the second expects to import one child row at a time. I'd rather do a bigger refactor of this code later if that's ok?

apiv2/db_import/tests/populate_db.py
```diff
@@ -69,227 +67,278 @@ def stale_deposition_author() -> dict:
...
+        if obj:
+            session.add(obj)
+        session.flush()
+    return wrapper
+
 
-def populate_deposition_authors() -> None:
-    DepositionAuthor.create(**stale_deposition_author())
+@write_data
+def populate_deposition(session: sa.orm.Session) -> Deposition:
```
Bento007, line 81: 1. I'm assuming the order you call these `populate` functions matters since it's building relationships within the database. If the relationships are not setup correctly I image the commit with fail right? That requirement is not clear from reading the code alone. We could either point the user to the documentation of the database schema or code the ordering in additional populate functions. I prefer coding it.

2. It would be easier to manage if we made the `populate` functions pytest fixtures. The only parameter they use is `session` which is already a pytest fixture, and the data is hardcoded.

3.  We can also simplify some of the tests a little implement both 1 and 2 by commiting the session in the fixture. Then we could remove `sync_db_session` from most of the tests.

apiv2/db_import/tests/test_db_dataset_import.py
```diff
@@ -74,74 +76,82 @@ def expected_funding_sources() -> list[dict[str, Any]]:
...
+    populate_dataset(sync_db_session)
+    populate_dataset_authors(sync_db_session)
+    populate_dataset_funding(sync_db_session)
+    sync_db_session.commit()
+    actual = verify_dataset_import()
     assert len(actual.authors) == 1
     assert len(actual.funding_sources) == 1
 
 
 # Tests addition of new authors, updating entries already existing in db, and removing stale authors
```
Bento007, line 98: pytest will the show the docstring for the test if it has one. Consider moving these comments to the docstring.

apiv2/template_overrides/support/enums.py.j2
```diff
@@ -0,0 +1,19 @@
+"""
```
Bento007, line 1: should this file be to `.gitattributes`?

Bento007 (changes requested):
1. where does the database get cleanup in-between tests?
2. should *apiv2/schema/schema.yaml* and *api/v2.0.0/api_models.yaml* be consolidated?

jgadling:
> 1. where does the database get cleanup in-between tests?

There's fancy magic in `apiv2/conftest.py` that uses [pytest_postgresql's DatabaseJanitor](https://pypi.org/project/pytest-postgresql/) to automatically create test databases for each test and drop the database at the end of each test

>     2. should _apiv2/schema/schema.yaml_ and _api/v2.0.0/api_models.yaml_ be consolidated?

Yes, they need to be consolidated 💯 but that's a heavy lift right now - we need to make changes to `schema.py` (and incorporate some upstream fixes in the LinkML library!) to do that properly, so for now we have two separate schema files as a kludge.

apiv2/db_import/importers/deposition.py
```diff
@@ -28,20 +32,20 @@ def get_id_fields(cls) -> list[str]:
...
+    def get_db_model_class(cls) -> type[models.Deposition]:
+        return models.Deposition
 
     @classmethod
     def get_direct_mapped_fields(cls) -> dict[str, Any]:
         return {
             "id": ["deposition_identifier"],
-            "title": ["deposition_title"],
-            "description": ["deposition_description"],
+            "deposition_title": ["deposition_title"],
```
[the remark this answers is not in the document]
jgadling: I don't want to get into too many discussions about field names / schema stuff in this PR -- there's another PR (#277) that ensures that the APIv2 schema is as in-sync as possible with the V1 API, even when the V1 api is weirdly inconsistent about field naming - we've decided that we want maximum compatibility with the old schema vs keeping field names consistent.

apiv2/db_import/importers/dataset.py
```diff
@@ -1,14 +1,15 @@
 from typing import Any, Iterable
 
-import importers.db.deposition
-from common import db_models
-from common.db_models import BaseModel
-from importers.db.base_importer import (
+from database import models
```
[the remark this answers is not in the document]
jgadling: I generally avoid aliasing imported modules unless there's a conflict, or the module has a well-known convention for aliasing (like pandas and sqlalchemy) - do you think that reducing lines of code changed in the PR is worth the tradeoff of remembering to use the alias when writing code?

apiv2/db_import/tests/populate_db.py
```diff
@@ -0,0 +1,537 @@
...
+    return Deposition(**stale_deposition_metadata())
+
+
+@write_data
+def populate_deposition_authors(session: sa.orm.Session) -> DepositionAuthor:
+    return DepositionAuthor(**stale_deposition_author())
+
+
+@write_data
+def populate_dataset(session: sa.orm.Session) -> Dataset:
```
jgadling, line 91: TODO: Make sure that the `Dataset.deposition_id` and `Tomogram.deposition_id` fields are not nullable, and update the tests to create depositions automatically when we create datasets, so that we do *not* need to run these `populate_xxxx` methods in a special order.

apiv2/db_import/tests/populate_db.py
```diff
@@ -0,0 +1,537 @@
...
+
+
+@write_data
+def populate_dataset_authors(session: sa.orm.Session) -> DatasetAuthor:
+    return DatasetAuthor(id=DATASET_AUTHOR_ID, dataset_id=DATASET_ID, name="Virginia Woolf", author_list_order=3)
+
+
+@write_data
+def populate_stale_dataset_authors(session: sa.orm.Session) -> DatasetAuthor:
+    return DatasetAuthor(dataset_id=DATASET_ID, name="Stale Author", author_list_order=1)
```
jgadling, line 114: TODO: review how these methods get used in tests and figure out why they don't depend on a dataset being created and at *least* add a comment about why they're weird.

apiv2/db_import/importer.py
```diff
@@ -0,0 +1,246 @@
...
+                voxel_spacing_cleaner.mark_as_active(voxel_spacing_obj)
+
+            voxel_spacing_cleaner.remove_stale_objects()
+
+        run_cleaner.remove_stale_objects()
+        session.commit()
+    session.commit()
+
+
+if __name__ == "__main__":
```
manasaV3, line 245: nit: is this needed?

apiv2/db_import/tests/populate_db.py
Bento007: will be resolved in https://github.com/chanzuckerberg/cryoet-portal/issues/121

apiv2/db_import/importers/annotation.py
```diff
@@ -0,0 +1,237 @@
...
+        annotation_shapes = AnnotationShapesDBImporter.get_item(annotation_obj.id, self, self.config)
+        annotation_shape = annotation_shapes.import_to_db()
+        annotation_files = AnnotationFilesDBImporter.get_item(
+            annotation_shape.id,
+            self.voxel_spacing_id,
+            self,
+            self.config,
+        )
+        annotation_files.import_to_db()
+        return annotation_obj
```
manasaV3, line 73: nit: Should this be in the `AnnotationShapesDBImporter.import_to_db()`?

apiv2/db_import/importers/annotation.py
```diff
@@ -0,0 +1,237 @@
...
+            "affiliation_name": ["affiliation_name"],
+            "affiliation_address": ["affiliation_address"],
+            "affiliation_identifier": ["affiliation_identifier"],
+            "author_list_order": ["author_list_order"],
+        }
+
+    def update_data_map(self, data_map: dict[str, Any], metadata: dict[str, Any], index: int) -> dict[str, Any]:
+        data_map = super().update_data_map(data_map, metadata, index)
+        primary_author_status = {
+            "primary_annotator_status": metadata.get("primary_annotator_status", metadata.get("primary_author_status")),
```
manasaV3, line 204: Could we use this migration as an opportunity to deprecate `primary_annotator_status` in favour of `primary_author_status`?

manasaV3 (approved):

apiv2/db_import/importer.py
jgadling: Yep, it allows us to run the script on the command line or also import its methods into another module

PR #276, opened 2024-09-18
feat: Adding destination filter finder

manasaV3:
Relates to https://github.com/chanzuckerberg/cryoet-data-portal/issues/1053

## Description
Creates a finder for adding relationships based on the fields in the metadata.

### Changes:
- Moving the resolve output paths to the relevant classes
- Adds check to prevent destination finder to reimport data
- Creates new finder `destination_filter` to find existing files using the metadata path
- Adds a relationship between tiltseries and alignments


ingestion_tools/scripts/standardize_dirs.py
```diff
@@ -126,7 +62,8 @@ def do_import(config, tree, to_import, metadata_import, to_iterate, kwargs, pare
             if filter_patterns and not list(filter(lambda x: x.match(item.name), filter_patterns)):
                 print(f"Skipping {item.name}..")
                 continue
-            if import_class in to_import:
+            # If item is to be imported, only import it if the source data allows import
+            if import_class in to_import and item.allow_imports:
```
jgadling, line 66: I think having the `allow_imports` checks in the `import_item()` and `import_metadata()` methods makes it harder to accidentally import data when we shouldn't. Relying on the caller to check this flag seems pretty insecure to me.

ingestion_tools/scripts/importers/base_importer.py
```diff
@@ -126,6 +133,19 @@ def get_default_config(cls) -> list[dict] | None:
...
 
+    @classmethod
+    def get_name_and_path(cls, metadata: dict, name: str, path: str, results: dict[str, str]) -> [str, str, dict]:
+        """
+        Returns the name, path and a dictionary of name and paths for the importer. This method is used to override the
+        name and path for the importer when the destination metadata is provided.
+        :param metadata: the metadata associated to the relevant importer entity
+        :param name: the name identified for the importer entity
+        :param path: the path for the importer entity
+        :param results: a dict of the filename and path for the importer entity
```
jgadling, line 144: this `results` parameter is unused so far - why do we need it?

manasaV3: This will be used in cases like Alignment, where we might want to merge the data from the finder with data from the metadata.

ingestion_tools/scripts/common/finders.py
```diff
@@ -105,12 +108,47 @@ class BaseLiteralValueFinder(BaseFinder):
...
         self.literal_value = value
 
-    def find(self, config: DepositionImportConfig, glob_vars: dict[str, Any]) -> dict[str, str | None]:
+    def find(self, config: DepositionImportConfig, glob_vars: dict[str, Any], *args, **kwargs) -> dict[str, str | None]:
         if isinstance(self.literal_value, dict):
             return self.literal_value
         return {item: None for item in self.literal_value}
 
 
+class DestinationFilteredMetadataFinder(BaseFinder):
```
jgadling, line 117: Tests?

ingestion_tools/scripts/common/finders.py
```diff
@@ -105,12 +108,47 @@ class BaseLiteralValueFinder(BaseFinder):
...
+            if isinstance(value, dict):
+                value = value.get(path_part)
+            else:
+                value = {v.get(path_part) for v in value if isinstance(v, dict)}
+            if not value:
+                break
+
+        return expected_value in value if isinstance(value, set) else value == expected_value
+
+    def find(self, config: DepositionImportConfig, glob_vars: dict[str, Any], *args, **kwargs) -> dict[str, str | None]:
```
jgadling, line 138: Can `cls` be a required keyword parameter? I'm actually wondering if the importer class should be a required parameter for the finder constructor instead of a parameter on `find`

ingestion_tools/scripts/common/finders.py
```diff
@@ -217,6 +265,22 @@ def find(
             return []
         return self._get_results(cls, config, metadata, **parent_objects)
 
+    @classmethod
+    def handle_for_metadata_file(
+        cls,
+        klass,
```
Bento007, line 271: what is klass, i'm not familiar with this notation?

Bento007 (commented):
Thank for simplifying the paths.

ingestion_tools/scripts/common/finders.py
manasaV3: Tests added

ingestion_tools/scripts/common/finders.py
manasaV3: The `klass` was supposed to be the Importer class. I will document that in code as well. :)

Bento007 (approved):

ingestion_tools/scripts/common/config.py
```diff
@@ -221,37 +221,11 @@ def get_metadata_path(self, obj: BaseImporter) -> str:
...
-            "deposition": "depositions_metadata/{deposition_name}",
-            "deposition_metadata": "depositions_metadata/{deposition_name}/deposition_metadata.json",
-            "deposition_keyphoto": "depositions_metadata/{deposition_name}/Images",
-        }
+        from importers.utils import get_importer_output_path
 
         output_prefix = self.output_prefix
         glob_vars = obj.get_glob_vars()
-        path = os.path.join(output_prefix, paths[key].format(**glob_vars))
+        path = os.path.join(output_prefix, get_importer_output_path(key).format(**glob_vars))
```
jgadling, line 228: tiny nit -- I think `get_importer_output_path` should handle expanding the glob vars

ingestion_tools/scripts/common/finders.py
```diff
@@ -249,22 +302,25 @@ class MultiSourceFileFinder(DefaultImporterFactory):
...
         config: DepositionImportConfig,
         metadata: dict[str, Any],
         filtered_results: dict[str, str],
+        allow_imports: bool,
         parents: dict[str, Any] | None,
     ):
         if not filtered_results:
             return []
+        if any(path and path.endswith("metadata.json") for path in filtered_results.values()):
+            name, path, filtered_results = self.importer_cls.get_name_and_path(metadata, None, None, filtered_results)
```
jgadling, line 314: `get_name_and_path` is only implemented for the TiltSeriesImporter - are we never going to wind up in this situation for any other importers?

ingestion_tools/scripts/importers/alignment.py
```diff
@@ -50,30 +52,27 @@ class AlignmentImporter(BaseFileImporter):
...
-        metadata: dict[str, Any],
-        name: str,
-        path: str,
-        parents: dict[str, Any],
-        file_paths: dict[str, str],
-    ):
-        super().__init__(config, metadata, name, path, parents)
-        self.identifier = AlignmentIdentifierHelper.get_identifier(config, metadata, parents)
-        self.file_paths = file_paths
+    def __init__(self, *args, **kwargs):
```
jgadling, line 60:
```suggestion
    def __init__(self, *args, file_paths: dict[str, str], **kwargs):
```

ingestion_tools/scripts/importers/alignment.py
```diff
@@ -50,30 +52,27 @@ class AlignmentImporter(BaseFileImporter):
...
+        super().__init__(*args, **kwargs)
+        self.file_paths = kwargs.get("file_paths")
```
jgadling, line 62:
```suggestion
        self.file_paths = file_paths
```

ingestion_tools/scripts/common/finders.py
```diff
@@ -185,45 +226,55 @@ def _get_results(
...
         metadata: dict[str, Any],
         filtered_results: dict[str, str],
+        allow_imports: bool,
         parents: dict[str, Any] | None,
     ) -> list[BaseImporter]:
         results = []
         for name, path in filtered_results.items():
-            item = self._instantiate(cls, config, metadata, name, path, parents)
+            name, path, metadata = self.handle_for_metadata_file(config, name, path, metadata)
+            item = self._instantiate(config, metadata, name, path, allow_imports, parents)
```
jgadling, line 245: Again with special casing for the `*_metadata.json` files. Can we at least get some comments (and tests!) that cover this special case?

jgadling (approved):
I think we need to document how the metadata file handling for the finders works somewhere (such as when we'd want to use it, why we treat that case differently, and which importers support it)  - either as code comments, or in the docs for the ingestion configs. I'm approving it so that this can be merged once those concerns are addressed

ingestion_tools/scripts/common/finders.py
manasaV3: This is going to be implemented only for alignment and tomogram for now. This will be added in a future PR, where those relationships are added.

jgadling (approved):

PR #277, opened 2024-09-19
feat: Align v1 and v2 schemas better

jgadling:
This PR tries to align the v1 and v2 api schemas better - reusing field names and types where possible, even if it makes maintaining our linkml schema a little bit less orderly. For example: we have `Deposition.deposition_publications` and `Dataset.dataset_publications` fields, when both `Deposition` and `Dataset` could inherit a `publications` field from a mixin - but we're preferring to keep backwards compatibility in this case.

I also tried to keep the descriptions for each field in sync with the values in apiv1 -- these descriptions ultimately wind up in the documentation for the API Client!

The main changes in this PR are in `schema/schema.yaml` with a few related changes to `db_import/**` and `scrape.py` to support field renaming.


apiv2/schema/schema.yaml
```diff
@@ -699,25 +677,36 @@ classes:
...
     - AuthorEntityMixin
     attributes:
       annotation:
         name: annotation
-        description: Metadata about an annotation for a run
+        description: Reference to the annotation this author contributed to
         from_schema: cdp-dataset-config
         inverse: Annotation.authors
         range: Annotation
+    slot_usage:
```
jgadling, line 692: We're inheriting a bunch of fields from the `AuthorEntityMixin` and this `slot_usage` block lets us override properties of the inherited fields

jgadling:
NOTE: we should also handle for case where an annotation submitted by the dataset_author is the portal_standard.

jgadling:
NOTE: need to add support for a `simulated` annotation method

jgadling:
> NOTE: need to add support for a `simulated` annotation method

apiv2/database/models/dataset.py
```diff
@@ -81,8 +81,8 @@ class Dataset(Base):
     deposition_date: Mapped[datetime.datetime] = mapped_column(DateTime(timezone=True), nullable=False)
     release_date: Mapped[datetime.datetime] = mapped_column(DateTime(timezone=True), nullable=False)
     last_modified_date: Mapped[datetime.datetime] = mapped_column(DateTime(timezone=True), nullable=False)
-    publications: Mapped[str] = mapped_column(String, nullable=True)
+    dataset_publications: Mapped[str] = mapped_column(String, nullable=True)
```
manasaV3, line 84: What do you think of having this field be `publications` across the dataset, deposition and tomogram table as the prefix is redundant? 
We could use the `CrossReferenceMixin` to keep this field and the `related_database_entries` consistent.

apiv2/database/models/tomogram.py
```diff
@@ -72,9 +72,9 @@ class Tomogram(Base):
         back_populates="tomograms",
     )
     name: Mapped[str] = mapped_column(String, nullable=True)
-    size_x: Mapped[float] = mapped_column(Float, nullable=False)
-    size_y: Mapped[float] = mapped_column(Float, nullable=False)
-    size_z: Mapped[float] = mapped_column(Float, nullable=False)
```
manasaV3, line 77: We should add `is_visualization_default`, `publications` and `related_database_entries` fields to this table as well.

apiv2/schema/schema.yaml
```diff
@@ -816,7 +809,7 @@ classes:
       plural:
         tag: plural
         value: Annotations
-    description: Metadata about an annotation for a run
```
manasaV3, line 819: I think this table is missing the `method_links` json field.

apiv2/database/models/tomogram.py
manasaV3: Resolving it here, as I added this [comment to the other PR.](https://github.com/chanzuckerberg/cryoet-data-portal-backend/pull/280#discussion_r1774575521)

apiv2/database/models/dataset.py
jgadling: Basically this PR is specifically to try to keep as much compatibility as possible with the V1 api, which was the preference that CZII had for the migration, and in the old api, the field was `dataset_publications`

apiv2/schema/schema.yaml
jgadling: Added!

apiv2/schema/schema.yaml
```diff
@@ -887,6 +879,9 @@ classes:
         - annotation_method
         range: string
         required: true
+      method_links:
+        description: Provides links to information on the method used for generating annotation, comma separated
+        range: string
```
manasaV3, line 884: I believe this is currently a json field.

manasaV3 (approved):

PR #278, opened 2024-09-19
fix: make sure bool comparators don't require int input.

jgadling:
The Frrontend team noticed that query filters for boolean values were erroneously requiring integer values as input. This updates the GraphQL interface to require bools as input instead. I also removed the comparisons (gt, lt) that looked irrelevant to boolean comparisons
Screenshot 2024-09-19 at 12 51 07 PM


bchu1 (approved):
Thanks!

manasaV3 (approved):

PR #280, opened 2024-09-19
feat: track gain and mdoc files alongside runs

jgadling:
This is a draft to make sure we're aligned on the structure (and naming!) for keeping track of mdoc and gains files associated with a run. Once we agree on names/structures, I'll update the API with the changes.

Also feel free to just add commits to this branch if that's easier.


anchi2c (commented):
What happened to per_section_parameters ? When I look at the changes, they are removed in the new changes.

jgadling:
> What happened to per_section_parameters ? When I look at the changes, they are removed in the new changes.

We're going to remove them for now because they'll be coming in a later update. It wouldn't hurt to just include them early, *except* that I'm worried we'll be making breaking changes (like removing/renaming fields in these tables) once we start working with that data, so in that case it would be better to add them in once we're more confident that the structure of those tables is stable.

apiv2/schema/schema.yaml
```diff
@@ -1545,146 +1582,6 @@ classes:
...
-        from_schema: cdp-dataset-config
-        exact_mappings:
-        - per_section_astigmatic_angle
-        range: float
-        minimum_value: -180
-        maximum_value: 180
-        unit:
-          symbol: °
-          descriptive_name: degrees
-  PerSectionAlignmentParameters:
```
manasaV3, line 1611: I believe we still want to retain the `PerSectionAlignmentParameters`?

apiv2/platformics/graphql_api/core/deps.py
```diff
@@ -38,7 +38,11 @@ async def get_db_session(
     try:
         yield session
     finally:
+        print("===== ")
+        print("===== CLOSING SESSION =====")
+        print("===== ")
         await session.close()  # type: ignore
+    print("===== CLOSING SESSION =====")
```
manasaV3, line 45: Do we want to retain all of these print statements?

apiv2/schema/schema.yaml
```diff
@@ -1723,6 +1696,14 @@ classes:
         multivalued: true
         inverse: Frame.run
         range: Frame
+      gain_files:
```
manasaV3, line 1699: Could we also update the description of the `run.dataset` in [here](https://github.com/chanzuckerberg/cryoet-data-portal-backend/pull/280/files#diff-145ab66c6e4310481f4a1eaf8c311191a0349f39f7d7d9ac92207ecdfda3a6c4R1688)?

apiv2/schema/schema.yaml
```diff
@@ -2322,11 +2261,14 @@ classes:
...
-        description: whether this tomogram is canonical for the run
-        from_schema: cdp-dataset-config
+      is_portal_standard:
+        description: whether this tomogram adheres to portal standards
         range: boolean
+        ifabsent: false
+      is_author_submitted:
+        description: whether this tomogram was submitted by the author of the dataset it belongs to.
+        range: boolean
+        ifabsent: false
```
manasaV3, line 2271: I believe this table also needs `is_visualization_default`, `publications` and `related_database_entries` fields.

apiv2/schema/schema.yaml
```diff
@@ -1545,69 +1581,6 @@ classes:
...
-        exact_mappings:
-        - per_section_astigmatic_angle
-        range: float
-        minimum_value: -180
-        maximum_value: 180
-        unit:
-          symbol: °
-          descriptive_name: degrees
   PerSectionAlignmentParameters:
     name: PerSectionAlignmentParameters
```
manasaV3, line 1585: - The `in_plane_rotation` field will be a collection of 4 float values. Should we make it a vector or a string? 
- The `beam_tilt` field is not going to be supported for now. 
- We should add support for the `is_canonical` boolean field and the float `volume_x_rotation`.

apiv2/platformics/graphql_api/core/deps.py
jgadling: I mangled this PR, sorry.

apiv2/schema/schema.yaml
jgadling: what does`is_visualization_default` mean for a tomogram? is that the default for an alignment or a run?

apiv2/schema/schema.yaml
jgadling: Supporting vectors and other *strongly typed* json fields is pretty doable in the API - we already support 2d arays of floats, so supporting 1d arrays is easy to handle. I haven't tried to add complex types to the client before though, so that could get interesting. Do you think `in_plane_rotation` would be best represented as a list of floats?

manasaV3: @uermel Could we have your thoughts on what is the best way to represent `in_plane_rotation` for the end user?

manasaV3 (approved):

apiv2/schema/schema.yaml
jgadling: Decision: This will be a 2d rotation matrix, basically `list[list[float]]`

PR #281, opened 2024-09-20
fix: Update sync script to sync missing fields.

jgadling:
The frontend team noticed some data missing from the v2 api. I wrote a script to check for empty db columns, and updated the sync script to make sure we're copying over the missing data.


bchu1 (approved):

bchu1:
Thanks!

PR #283, opened 2024-09-24
feat: Minor tomogram table update

manasaV3:
Relates to

## Description
- ~~Adds is_author_submitted field to tomogram~~
- Removes is_canonical field to tomogram


apiv2/schema/schema.yaml
```diff
@@ -2441,7 +2436,13 @@ classes:
         range: boolean
         required: true
         # TODO - is this reasonable?
-        ifabsent: false
+        ifabsent: falsegit
```
jgadling, line 2439: typo here I think

jgadling:
I think you need to run `make update-schema` and/or add the generated migration file to this PR.

PR #285, opened 2024-09-24
fix: Make sure we're using DB connection pools properly.

jgadling:
This fixes two major bugs in apiv2:
1. The api was creating a new database connection pool on every request, which basically meant that we had no limit on the number of db connections between the app and the database. We're now creating the pool once at the startup of the app, and then fetching connections from that pool in the rest of the app. I've currently set the limit at 5 connections (which can burst up to 10 connections) per worker. Each pod has 4 workers, and we have two pods running in most environments, which means a max of 80 connections possible from the api, while the limit at the database server is around 200 connections, so we should be in good shape for awhile.
2. We weren't properly closing database sessions due to the way we were handling our integration with FastAPI's dependency injection system -- we needed to be using [AsyncExitStack](https://docs.python.org/3/library/contextlib.html#contextlib.AsyncExitStack) as a context manager to make sure any fastapi dependencies that `yield` control ([docs/examples here](https://fastapi.tiangolo.com/tutorial/dependencies/dependencies-with-yield/)) had control returned back to them after they'd yielded.

### DIFF
--- a/apiv2/main.py
+++ b/apiv2/main.py
@@ -9,8 +9,13 @@ import uvicorn
 from graphql_api.mutations import Mutation
 from graphql_api.queries import Query
 from starlette.middleware.cors import CORSMiddleware
+from starlette.requests import Request
 
-from platformics.graphql_api.core.deps import get_auth_principal
+from platformics.database.connect import AsyncDB, init_async_db
+from platformics.graphql_api.core.deps import (
+    get_auth_principal,
+    get_engine,
+)
 from platformics.graphql_api.core.error_handler import HandleErrors
 from platformics.graphql_api.setup import get_app, get_strawberry_config
 from platformics.security.authorization import Principal
@@ -22,6 +27,9 @@ schema = strawberry.Schema(query=Query, mutation=Mutation, config=get_strawberry
 
 # Create and run app
 app = get_app(settings, schema)
+
+engine = init_async_db(settings.DB_URI, echo=settings.DB_ECHO, pool_size=5, max_overflow=5)
+app.state.engine = engine
 
 
 def get_allowed_origins() -> list[str] | None:
@@ -64,6 +72,11 @@ def override_auth_principal():
     )
 
 
+def override_get_engine(request: Request) -> AsyncDB:
+    return request.app.state.engine
+
+
+app.dependency_overrides[get_engine] = override_get_engine
 app.dependency_overrides[get_auth_principal] = override_auth_principal
 
 if __name__ == "__main__":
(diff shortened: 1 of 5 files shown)


apiv2/main.py
```diff
@@ -23,6 +28,9 @@
 # Create and run app
 app = get_app(settings, schema)
 
+engine = init_async_db(settings.DB_URI, echo=settings.DB_ECHO, pool_size=5, max_overflow=5)
+app.state.engine = engine
```
jgadling, line 32: Creates the engine once and stashes it somewhere we can fetch it later

apiv2/main.py
```diff
@@ -64,6 +72,11 @@ def override_auth_principal():
     )
 
 
+def override_get_engine(request: Request) -> AsyncDB:
+    return request.app.state.engine
```
jgadling, line 76: Where we fetch the engine later.

apiv2/platformics/database/connect.py
```diff
@@ -40,7 +40,7 @@ def session(self) -> sessionmaker[Session]:
 
 
 def init_async_db(db_uri: str, **kwargs: dict[str, Any]) -> AsyncDB:
-    engine = create_async_engine(db_uri, pool_size=5, max_overflow=5, future=True, **kwargs)
+    engine = create_async_engine(db_uri, future=True, **kwargs)
```
jgadling, line 43: We can pass these in as kwargs

apiv2/platformics/graphql_api/core/strawberry_extensions.py
```diff
@@ -61,27 +61,22 @@ async def resolve_async(
         except AttributeError:
             request.context = {"dependency_cache": {}}
 
-        if not request.scope.get("sberrystack"):
+        async with AsyncExitStack() as async_exit_stack:
```
jgadling, line 64: How to use `AsyncExitStack` properly

apiv2/platformics/graphql_api/core/strawberry_extensions.py
```diff
@@ -61,27 +61,22 @@ async def resolve_async(
...
+                request=request,
+                dependant=self.dependant,
+                body={},
+                dependency_overrides_provider=request.app,
+                dependency_cache=request.context["dependency_cache"],
+                async_exit_stack=async_exit_stack,
+                embed_body_fields=False,
+            )
+            solved_values = solved_result.values
+            new_cache = solved_result.dependency_cache
```
jgadling, line 77: I upgraded fastapi as part of this quest, and `solved_result` changed its shape in the new version

bchu1 (approved):

bchu1:
Thanks! 🙌

PR #262, opened 2024-09-10
feat: Update ingestion config validation for ctf and alignment entities

Bento007:
related to https://github.com/chanzuckerberg/cryoet-data-portal/issues/997

### update existing ingest configs
- added a migrate script to migrate existing ingest_configs to the new schema. This cause some reformatting of spaces and newlines of unrelated fields.
- removed fields that were empty lists or empty mappings except for `annotations`

#### collection_metadata 

- move `*.mdocs` from `rawtilts[].sources` to `collection_metadata[].sources[].source_multi_glob.list_globs`

#### alignments

- copy`tomogram[0].metadata.affine_transformation_matrix` to `alignment[].metadata.affine_transformation_matrix` if it is not an identity matrix, and if alignment sources existed. `affine_transformation_matrix` was copied into each `alignment[].source`
- add `alignment[].metadata.format` and set to appropriate value based on source formats.
- - move `*.tlt`, `*.xf`, `*.com`, and `*.aln`   from `rawtilts[].sources` to `collection_metadata[].sources[].source_multi_glob.list_globs` 

#### tomograms
- remove `tomogram[].metadata.affine_transformation_matrix` if it is an identity matrix. 
- add `tomogram[].metadata.dates` from `deposition[0].metadata.dates` **if it existed**, else the **epoch date** was used for all dates. Here is a list of the configs a default date was used 
 - 10011_draft.yaml 
 - ~10431.yaml~
  - ~10001.yaml~
  - ~10002.yaml~
  - ~10302.yaml~
  - ~10427.yaml~
  - ~10428.yam~
  - ~10429.yaml~
  - ~10430.yaml~
  - ~10432.yaml~
  - ~10434.yaml~
  - ~10435.yaml~
  - ~10437.yaml~
  - ~10438.yaml~
- add `tomogram[].metadata.is_visualization_default` with default set to `true`

### Template.yaml
- add `alignments`
  - alignment_type
  - offset
  - x_roation_offset
  - tilt_offset
  -  affine_transformation_matrix
  - is_canonical
  - format
  - volume_dimesion
- add `annotations[].sources.*.is_portal_standard`
- add `frames[].metadata`
  - dose
  - defocus
  - astigmatism
  - astigmatic_angle
- add `tomograms[].metadata`
  - `is_visualization_default`
  - `is_portal_standard`
  - `cross_references`
  - `dates`

### Ingestion Config Schema
- define alignment metadata and sources and add it to ingest_config_models.yaml
[...]

PR #290, opened 2024-09-26
feat: transform_ingestion_configs.py support multiple versions

Bento007:
- Add a docstring and description for the upgrade command
- Update transform_ingestion_configs to be extended to support upgrading an ingest config from version 0.0.0 and beyond.
- move schema migrations code to ingest_tools/schema_migration
- update gjensen_config.py to run with reorganization.
- related to #262 

# Testing

Test manually by modifying an existing ingest config to a previous version.

### DIFF
--- a/ingestion_tools/scripts/transform_ingestion_configs.py
+++ b/ingestion_tools/scripts/transform_ingestion_configs.py
@@ -1,3 +1,4 @@
+from collections import OrderedDict
 from typing import Any
 
 import click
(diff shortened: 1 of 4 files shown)


jgadling (approved):
I think we need to get to the point where we have a tool that knows how to run ingestion config migrations relative to the current version of the config files, but this provides some value for now.

ingestion_tools/scripts/transform_ingestion_configs.py
```diff
@@ -35,8 +36,9 @@ def has_no_sources(data: list[dict[str, Any]] | dict[str, Any]) -> bool:
     return isinstance(data, dict) or not any(row.get("sources") for row in data)
 
 
-def update_config(data: dict[str, Any]) -> dict[str, Any]:
+def update_config_to_v1(data: dict[str, Any]) -> dict[str, Any]:
```
jgadling, line 39: TODO - changing this method name breaks the gjensen transformation script.

jgadling: TODO:
- Make sure that this code gets executed by `gjensen_config.py` to make sure gjensen outputs get written as v1.1
- Move these ingestion configs manipulation scripts to `ingestion_tools/scripts/migrations`

PR #292, opened 2024-09-26
chore: reorganize ingest_tools/schema_migration

Bento007:
## Description
- move version migrations to their own file
- change update to upgrade
- move cli configuration to its own file
- remove unused convert_version function
- move issue_997.py to ingestion-tools/schema-migration and renamed to migrate_v1_1_0.py


ingestion_tools/scripts/schema_migration/upgrade.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def upgrade_config(data: dict[str, Any]) -> dict[str, Any]:
+    version_map = OrderedDict(
+        # Version_map must be updated when a new migration is needed.
+        # Order matters.
+        {
+            # current_version: (update_function, next_version)
+            "0.0.0": (migrate_v1_0_0.upgrade, "1.0.0"),
+            "1.0.0": (migrate_v1_1_0.upgrade, "1.1.0")})
```
jgadling, line 20: NIT: it'd be nice to get to the point where the script is smart enough to read the list of migrations from a directory and check the version tags in each config file against the list of migrations available, and then runs only the necessary migrations (more like the way alembic works) but this is a good start!

jgadling (approved):

manasaV3 (approved):

PR #296, opened 2024-10-01
fix: support filtering entities by related object id's

jgadling:
This PR does a couple things:
1. It removes `lib` from `.gitignore` since this was preventing us from checkin in an important file!
2. Adds that important file to the repo, kindof belatedly.
3. Updates codegen to add `some_related_class_id` fields to where clauses so they can be filtered. This functionality is necessary to support the client!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,6 @@ dist/
 downloads/
 eggs/
 .eggs/
-lib/
 lib64/
 parts/
 sdist/
--- a/apiv2/platformics/codegen/lib/linkml_wrappers.py
+++ b/apiv2/platformics/codegen/lib/linkml_wrappers.py
@@ -1,0 +1,457 @@
+"""
+Helpers for writing Jinja2 templates based on LinkML schema objects.
+
+The wrapper classes in this module are entirely centered around providing convenience
+functions to keep complicated LinkML-specific logic out of our Jinja2 templates.
+"""
+
+import contextlib
+from functools import cached_property
+
+import strcase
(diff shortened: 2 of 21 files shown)


apiv2/platformics/codegen/lib/linkml_wrappers.py
```diff
@@ -0,0 +1,457 @@
+"""
```
jgadling, line 1: This file wasn't getting checked in previously, due to `lib` being in `.gitignore`! I'm checking it in now.

.gitignore
```diff
@@ -14,7 +14,6 @@ dist/
 downloads/
 eggs/
 .eggs/
-lib/
```
jgadling, line 17: Removing this so we can commit an important codegen module

apiv2/template_overrides/graphql_api/types/class_name.py.j2
```diff
@@ -282,6 +282,9 @@ class {{ cls.name }}WhereClause(TypedDict):
         {%- else %}
     {{ attr.name }}: Optional[{{ getComparators(attr) }}] | None
         {%- endif %}
+        {%- if attr.inverse and not attr.multivalued %}
+    {{ attr.name }}_id : Optional[{{ getComparators(attr.related_class.identifier) }}] | None
+        {%- endif %}
```
jgadling, line 287: Adds `some_class_id` fields to where clauses so we can filter on them.

manasaV3 (approved):

Bento007 (approved):

PR #297, opened 2024-10-01
chore: add seed script.

jgadling:
This is the start of a script that can seed our local dev environment with test data. It's not wired into anything right now, but it may be useful in the future for CI. It's basically a translation of [this script](https://github.com/chanzuckerberg/cryoet-data-portal/blob/main/client/python/cryoet_data_portal/tests/test_infra/hasura/migrations/cryoetdataportal/0000000000001_data/up.sql) to `factory_boy` with a few updates for our new schema (to support `alignments` and `annotation_shape` etc)


Bento007 (approved):
Very nice. Thank you for adding this

PR #303, opened 2024-10-07
fix: keep both `tomogram.is_canonical` and `tomogram.is_portal_standard`

jgadling:
to allow the frontend queries to migrate.

Relates to


apiv2/database/migrations/versions/20241002_152402_autogenerated.py
```diff
@@ -115,13 +115,16 @@ def upgrade() -> None:
     op.add_column("tomogram", sa.Column("deposition_date", sa.DateTime(timezone=True), nullable=True))
     op.add_column("tomogram", sa.Column("release_date", sa.DateTime(timezone=True), nullable=True))
     op.add_column("tomogram", sa.Column("last_modified_date", sa.DateTime(timezone=True), nullable=True))
-    op.drop_column("tomogram", "is_canonical")
+
+    conn = op.get_bind()
+    backfill_sql = sa.sql.text("UPDATE tomogram SET is_portal_standard = is_canonical")
+    conn.execute(backfill_sql)
```
jgadling, line 121: I'm modifying an *existing* migration because this migration hasn't been executed in prod yet. I put the `is_canonical` column back in place manually in staging.

apiv2/schema/schema.yaml
```diff
@@ -3,7 +3,7 @@ name: cdp-dataset-config
 description: Schema for dataset configs
 version: 1.1.0
 imports:
-- linkml:types
```
jgadling, line 6: my new editor config wants to indent lists now, and the community is waffly about which behavior is correct -- https://github.com/yaml/pyyaml/issues/234 😭

PR #308, opened 2024-10-10
feat: add frame & gain ingestion to db import

jgadling:
I also started working on a new interface for these db importers that more closely resembles the S3 importers, so hopefully they're easier to maintain going forward. I'll update the old importers to use the new base classes over the next few weeks.


apiv2/db_import/common/config.py
```diff
@@ -0,0 +1,103 @@
...
+
+if TYPE_CHECKING:
+    from mypy_boto3_s3 import S3Client
+else:
+    S3Client = object
+
+logger = logging.getLogger("config")
+
+
+class DBImportConfig:
```
jgadling, line 18: I moved this class out of `importers/base_importer.py` into its own file.

apiv2/db_import/common/finders.py
```diff
@@ -0,0 +1,45 @@
...
+    @abstractmethod
+    def __init__(self, config: DBImportConfig, **kwargs):
+        pass
+
+    @abstractmethod
+    def find(self, item_importer: ItemDBImporter) -> list[ItemDBImporter]:
+        pass
+
+
+class FileFinder(ItemFinder):
```
jgadling, line 29: We'll need to add more finders over time, but this is ~similar to the idea of finders in s3 ingestion.

apiv2/db_import/common/config.py
```diff
@@ -0,0 +1,103 @@
...
+                    self.s3_client.head_object(Bucket=self.bucket_name, Key=f"{subdir}{target_filename}")
+                    result.append(subdir)
+                except Exception:
+                    continue
+        return result
+
+    def glob_s3(self, prefix: str, glob_string: str, is_file: bool = True):
+        paginator = self.s3_client.get_paginator("list_objects_v2")
+        if prefix.startswith("s3://"):
+            prefix = "/".join(prefix.split("/")[3:])
```
jgadling, line 59: I added these two lines to this method to make it a little less fussy to work with.

apiv2/db_import/importers/frame.py
```diff
@@ -0,0 +1,53 @@
...
+    model_class = models.Frame
+
+    def load_computed_fields(self):
+        https_prefix = self.config.https_prefix
+        s3_prefix = self.config.s3_prefix
+
+        self.model_args["s3_frame_path"] = os.path.join(s3_prefix, self.input_data["file"])
+        self.model_args["https_frame_path"] = os.path.join(https_prefix, self.input_data["file"])
+        self.model_args["run_id"] = self.input_data["run"].id
+        self.model_args["deposition_id"] = self.input_data["deposition"].id
```
manasaV3, line 30: We currently don't have a metadata for this, so would we populating this from the dataset's deposition_id ?

manasaV3 (approved):

PR #310, opened 2024-10-11
fix: Update metadata fetch for MultiSourceFileFinder

manasaV3:
## Description
This fixes a bug for DestinationFinder.
Updates the metadata used by MultiSourceFileFinder to the remote metadata that lives in s3, instead of the metadata from ingestion config file.

### DIFF
--- a/ingestion_tools/scripts/common/finders.py
+++ b/ingestion_tools/scripts/common/finders.py
@@ -153,13 +153,18 @@ class DestinationFilteredMetadataFinder(BaseFinder):
                 metadata = json.load(metadata_file)
             if all(self._is_match(metadata, item.get("key"), item.get("value")) for item in self.filters):
                 responses[file_path] = file_path
-
         return responses
 
 
 ###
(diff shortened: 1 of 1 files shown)


ingestion_tools/scripts/common/finders.py
```diff
@@ -329,9 +332,13 @@ def _get_instantiated_results(
     ):
         if not filtered_results:
             return []
-        # If any of the files found are metadata files, we use the metadata file to instantiate the class
-        if any(path and path.endswith("metadata.json") for path in filtered_results.values()):
-            name, path, filtered_results = self.importer_cls.get_name_and_path(metadata, None, None, filtered_results)
+        # If the files found are metadata files, we use the metadata file to instantiate the class
+        if all(path and path.endswith("metadata.json") for path in filtered_results):
```
jgadling, line 336: are we supposed to be checking the keys or values of `filtered_results` here?

jgadling (approved):

PR #321, opened 2024-10-16
fix: handle formatted voxel spacing string in tomo metadata

manasaV3:
Relates to

## Description
Handles the voxel spacing being a formatted string in the ingestion config


jgadling (approved):

uermel (approved):
Fix works for me.

PR #322, opened 2024-10-16
chore: Removing invalid key_image config blocks

manasaV3:
Relates to: https://github.com/chanzuckerberg/cryoet-data-portal/issues/1249

## Description
In the past, we would always override the key_images config block with a default config block. This resulted in this invalid blocks not being identified earlier.


uermel (approved):
LGTM!

PR #323, opened 2024-10-16
fix: Updating alignment dimensions and offsets to Angstrom

manasaV3:
Relates to

so we have to specify the size independent of voxel spacing

### DIFF
--- a/ingestion_tools/scripts/importers/alignment.py
+++ b/ingestion_tools/scripts/importers/alignment.py
@@ -120,7 +120,8 @@ class AlignmentImporter(BaseFileImporter):
         if not tomogram:
             # If no source tomogram is found don't create a default alignment metadata file.
             raise TomogramNotFoundError()
-        return tomogram.get_source_volume_info().get_dimensions()
+        voxel_size = round(tomogram.get_source_volume_info().voxel_size, 3)
+        return {d: voxel_size for d in "xyz"}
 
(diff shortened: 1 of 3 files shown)


ingestion_tools/scripts/importers/alignment.py
```diff
@@ -120,7 +120,8 @@ def get_tomogram_volume_dimension(self) -> dict:
         if not tomogram:
             # If no source tomogram is found don't create a default alignment metadata file.
             raise TomogramNotFoundError()
-        return tomogram.get_source_volume_info().get_dimensions()
+        voxel_size = round(tomogram.get_source_volume_info().voxel_size, 3)
+        return {d: voxel_size for d in "xyz"}
```
uermel, line 124: The tomogram dimensions should not be equal to the voxelsize, but rather the pixel dimensions should be multiplied with the voxelsize to arrive at  the physical volume dimensions.

This should be pixel dimension * voxelsize, i.e.:

```
dim = tomogram.get_source_volume_info().get_dimensions()
voxel_size = round(tomogram.get_source_volume_info().voxel_size, 3)
return {d: voxel_size * dim[d] for d in "xyz"}

uermel (changes requested):
Voxel size needs to be multiplied by the dimension.

uermel (approved):
Looks good now!

jgadling (approved):

PR #330, opened 2024-10-18
fix: Handle no alignment in AreTomo Converter

manasaV3:
Relates to

## Description
Return empty list if no alignment file is available for AreTomo.


jgadling (approved):

uermel:
It surprises me that this case occurs. Is it the broken runs in 10302?

manasaV3:
> It surprises me that this case occurs. Is it the broken runs in 10302?
I noticed this for `10301` and `10438`.

PR #337, opened 2024-11-04
feat: Support importing points from copick

uermel:
Relates to N/A

## Description
Support importing `Point` and `OrientedPoint` from [copick](https://github.com/copick/copick) format.


manasaV3 (approved):
🚢

PR #339, opened 2024-11-04
chore: Handle no neuroglancer config file

manasaV3:
## Description

Sets the neuroglancer config to `None` for cases where there is no neuroglancer_config path specified in the tomogram metadata.


jgadling:
We'll want the equivalent change for the v2 db ingestion as well!

jgadling (approved):

PR #347, opened 2024-11-06
chore: Build an x86 version of the apiv2 image to execute in aws batch.

jgadling:
This updates our `push-local-ingestor-build` make target to build both an apiv2 image (for db ingestion) and an ingestion_tools image (for s3 ingestion and apiv1 db ingestion) and also updates the GitHub action that pushes our ingestion_tools docker image to build an apiv2 image as well.

It's important to note that we've taken on this extra step because we run db ingestion in an x86-only environment in AWS batch, while our apiv2 service runs on ARM. We're using python and libraries that generally aren't tied to CPU architecture, so I don't expect any major differences in behavior between the different images

### DIFF
--- a/.github/workflows/push-ingestor-build.yaml
+++ b/.github/workflows/push-ingestor-build.yaml
@@ -14,7 +14,7 @@ permissions:
   contents: read
 
 jobs:
-  deploy-to-staging:
+  push-ingestor-build:
     name: build and push ingestor to remote
     runs-on: ubuntu-latest
     environment: staging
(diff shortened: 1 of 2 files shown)


.github/workflows/push-ingestor-build.yaml
```diff
@@ -40,3 +40,29 @@ jobs:
           tags: ${{ secrets.ECR_REPO }}:${{ github.head_ref || github.ref_name }}
           push: true
           provenance: false
+  push-ingestorv2-build:
```
manasaV3, line 43: Should this be apiv2?

.github/workflows/push-ingestor-build.yaml
```diff
@@ -40,3 +40,29 @@ jobs:
...
+      - name: Login to Amazon ECR
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@v1
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Docker build & Push
+        uses: docker/build-push-action@v4
+        with:
+          context: ./apiv2
+          file: ./apiv2/Dockerfile
```
manasaV3, line 65: Not a blocker: It looks like, the only difference between the ingestion job and this one are these two values. What are your thoughts on having a composite action that can be used by both these jobs?

jgadling: I updated it to use a matrix build. check again?

manasaV3 (approved):
:shipit:

PR #350, opened 2024-11-06
fix: Adding is_visualization_default field to tomo ingestion

manasaV3:
Relates to

## Description

Updates mapping for the field of is_visualization_default.


jgadling (changes requested):
1. Need tests
2. I think there's another reference to `is_visualization_default` in this file that will override this change

jgadling (approved):

PR #352, opened 2024-11-06
feat: Generate tomogram key images with neuroglancer colors and including segmentation masks.

uermel:
Relates to N/A

## Description

The previous method of generating key images relied on a different colormap for generating key images than the neuroglancer visualization config. It also did not show segmentation masks at all. 

This PR:

1. Adds the ability to create key images including both, binary masks and points, to `generate_preview`. In case of points, the old method is used. In case of masks, a slab of the same size as in the tomogram is projected and rendered with transparency above the tomogram, but below the points. 

2. Separates out the color map generation step from the neuroglancer layer config generation step in the visualization config importer, so it can be re-used in other cases. Adds  a method `get_annotation_color` to the importer that returns a color given a combination of an annotation importer and a tomogram importer. 

3. Modifies the key image importer, so that colors are obtained from the visualization config importer using the method from step 2. Also adds the ability for the importer to find mask annotations and create `ZarrReader` instances for them.


ingestion_tools/scripts/importers/key_image.py
```diff
@@ -69,27 +70,53 @@ def get_existing_preview(self) -> tuple[np.ndarray | None, int | None]:
             return data, data.shape[-1]
         return None, None
 
-    def load_annotations(self) -> Generator[np.ndarray, None, None]:
+    def load_point_annotations(self) -> Generator[KeyPointAnnotation, None, None]:
+        vis_config_importer = VisualizationConfigImporter(self.config, {}, parents=self.parents)
         for annotation in AnnotationImporter.finder(self.config, **self.parents):
```
manasaV3, line 75: This will only fetch the annotations in the config.

ingestion_tools/scripts/importers/key_image.py
```diff
@@ -69,27 +70,53 @@ def get_existing_preview(self) -> tuple[np.ndarray | None, int | None]:
...
-                if annotation.shape.lower() == "point":
-                    break
-            except FileNotFoundError:
-                print(f"Unable to load annotation data for {annotation.get_output_filename(annotation_path)}")
+            is_default = annotation.is_visualization_default
+            if annotation.shape.lower() in {"orientedpoint", "point"} and is_default:
+                annotation_path = annotation.get_output_path()
+                try:
+                    annotation_data = annotation.get_output_data(annotation_path)
+                    color = vis_config_importer.get_annotation_color(annotation, self.get_tomogram())
```
manasaV3, line 81: Currently, the key photo gets generated before the neuroglancer config gets generated. We should update that ordering [here](https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/4cfb783d9ca707dc66766ee7a28ac2abddd517e1/ingestion_tools/scripts/importers/utils.py#L58-L59), to satisfy this dependency.

ingestion_tools/scripts/importers/key_image.py
uermel: I think that was the way it worked before as well.

manasaV3: Yes, it is. I called it out as there was an offline discussion about including annotations that aren't a part of the deposition in the key image. :)

manasaV3 (approved):

PR #356, opened 2024-11-13
fix: Support running v2 db import via batch job.

jgadling:
This adds a new WDL file that supports ingesting data into the v1 and v2 db's with the same `enqueue_runs.py` command. The docker image necessary to do this has already been added to our CI workflows.

NOTE: the WDL files have already been uploaded to s3, so this is ready to go.

### DIFF
--- a/ingestion_tools/infra/db_import-v0.0.2.wdl
+++ b/ingestion_tools/infra/db_import-v0.0.2.wdl
@@ -1,0 +1,114 @@
+version 1.0
+
+task cryoet_data_dbimport_workflow {
+    input {
+        String docker_image_id
+        String aws_region
+        String s3_bucket
+        String https_prefix
+        String environment
+        String flags
+    }
+
+    command <<<
+        set -euxo pipefail
+        export PYTHONUNBUFFERED=1
+        python --version 1>&2
+        ls -l 1>&2
+        original_dir=$(pwd)
+        cd /usr/src/app/ingestion_tools/scripts
+        set +x
+        POSTGRES_URL=$(aws secretsmanager get-secret-value --secret-id ~{environment}/db_uri | jq -r .SecretString | jq -r .db_uri)
+        echo python db_import.py load ~{s3_bucket} ~{https_prefix} POSTGRES_URL ~{flags} 1>&2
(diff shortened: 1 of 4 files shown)


ingestion_tools/infra/db_import-v0.0.2.wdl
```diff
@@ -0,0 +1,114 @@
...
+        aws_region = aws_region,
+        s3_bucket = s3_bucket,
+        https_prefix = https_prefix,
+        environment = environment,
+        flags = flags
+    }
+
+    call cryoet_data_dbimport_v2_workflow {
+        input:
+        done = cryoet_data_dbimport_workflow.done,
```
manasaV3, line 101: This is a cool way to maintain the dependency between the tasks.

manasaV3 (approved):

ingestion_tools/infra/db_import-v0.0.2.wdl
```diff
@@ -0,0 +1,114 @@
...
+        python --version 1>&2
+        ls -l 1>&2
+        pwd 1>&2
+        set +x
+        apt install -y curl unzip jq
+        curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" && unzip -q awscliv2.zip && ./aws/install && rm -rf ./aws
+        pip install psycopg2-binary
+        POSTGRES_URL=$(aws secretsmanager get-secret-value --secret-id ~{environment}/v2_db_uri | jq -r .SecretString | jq -r .db_uri)
+        # TODO - Scrape ID's from the v1 api so we keep things in sync
+        pip install cryoet-data-portal==3.1.1
```
manasaV3, line 60: Could we move this to the poetry dependencies?

jgadling: It's not a dependency for the API service... I'm on the fence right now about creating a separate Dockerfile for the ingestion image vs the API service image. Which is something I'll probably do in the next few weeks, since all of these installs in the batch jobs are getting silly.

ingestion_tools/infra/db_import-v0.0.2.wdl
jgadling: Yeah, it's a pretty normal WDL hack - we did something similar with CZGE.

PR #358, opened 2024-11-15
fix: Oriented point rendering in Neuroglancer

manasaV3:
Relates to https://github.com/chanzuckerberg/cryoet-data-portal/issues/1059

## Description

- Fixes the rendering of oriented point in neuroglancer.
- Adds tests for viz config generation

### DIFF
--- a/ingestion_tools/scripts/importers/visualization_config.py
+++ b/ingestion_tools/scripts/importers/visualization_config.py
@@ -90,18 +90,21 @@ class VisualizationConfigImporter(BaseImporter):
         name_prefix: str,
         color: str,
         resolution: tuple[float, float, float],
-        is_instance_segmentation: bool,
+        shape: str,
         **kwargs,
     ) -> dict[str, Any]:
-        return state_generator.generate_point_layer(
-            source_path,
-            f"{name_prefix} point",
-            self.config.https_prefix,
-            color=color,
-            scale=resolution,
-            is_visible=file_metadata.get("is_visualization_default"),
-            is_instance_segmentation=is_instance_segmentation,
-        )
+        args = {
+            "source": source_path,
+            "name": f"{name_prefix} point",
+            "url": self.config.https_prefix,
+            "color": color,
+            "scale": resolution,
+            "is_visible": file_metadata.get("is_visualization_default"),
+            "is_instance_segmentation": shape == "InstanceSegmentation",
+        }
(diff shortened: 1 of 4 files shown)


ingestion_tools/scripts/importers/visualization_config.py
```diff
@@ -90,18 +90,21 @@ def _to_point_layer(
...
-            f"{name_prefix} point",
-            self.config.https_prefix,
-            color=color,
-            scale=resolution,
-            is_visible=file_metadata.get("is_visualization_default"),
-            is_instance_segmentation=is_instance_segmentation,
-        )
+        args = {
+            "source": source_path,
+            "name": f"{name_prefix} point",
```
manasaV3, line 98: @uermel  As the point and the orientend point are going to be rendered differently, should we distinguish that in the name as well?

ingestion_tools/scripts/importers/visualization_config.py
```diff
@@ -90,18 +90,21 @@ def _to_point_layer(
...
+            "url": self.config.https_prefix,
+            "color": color,
+            "scale": resolution,
+            "is_visible": file_metadata.get("is_visualization_default"),
+            "is_instance_segmentation": shape == "InstanceSegmentation",
```
jgadling, line 103: nit: not sure if it's just me, but this is kindof ugly imop.

ingestion_tools/scripts/importers/visualization_config.py
uermel: I agree conceptually, but would the names become too long if we add 'oriented'?

jgadling (approved):

ingestion_tools/scripts/importers/visualization_config.py
uermel: On second thought 

`segmentation`
`orientedpoint`

have the same length, so I agree we should name properly.

uermel (approved):

ingestion_tools/scripts/importers/visualization_config.py
manasaV3: The variable name or doing the check against the shape value?

jgadling: setting the value for the `is_instance_segmentation` key with an inline boolean expression

PR #361, opened 2024-11-19
feat: Adding fallback sources when config defined sources don't exist

manasaV3:
Relates to: https://github.com/chanzuckerberg/cryoet-data-portal/issues/1337

## Description

- Removes the mutation to the object config with default config, and move the check for default config to the finder.
- As the key photo entities's sources are literals those are validated with fallback in the importer
- Increases test coverage


ingestion_tools/scripts/tests/fixtures/dataset/invalid_http_keyphoto.yaml
```diff
@@ -0,0 +1,33 @@
+dataset_keyphotos:
```
manasaV3, line 1: Todo: Update the tests to use the same config can be used with dynamic update to the key_photo entity.

uermel (approved):

ingestion_tools/scripts/importers/base_importer.py
```diff
@@ -117,13 +117,13 @@ def is_import_allowed(self):
...
-            for source in sources:
-                source_finder_factory = cls.finder_factory(source, cls)
-                for item in source_finder_factory.find(config, metadata, **parents):
-                    yield item
+        item_found = False
+        for item in cls._finder_execution(finder_configs, config, **parents):
+            item_found = True
+            yield item
+        if not item_found and (default_config := cls.get_default_config()):
+            for item in cls._finder_execution(default_config, config, **parents):
```
jgadling, line 125: I find it kind of strange that `cls.get_default_config()` doesn't kick in if the source config is *missing* from an ingestion config, but it only gets applied if the yaml source doesn't *find* anything. It's not a default configuration but instead it's a fallback if no results are found. It's a pretty subtle difference but I think it could be confusing to someone who's writing ingestion configs. Why do we want it to work this way? I think we should update the docstring for the `BaseImporter.get_default_config()` to explain how the method is really used and why.

manasaV3: If I am understanding this correctly, this fix should still hold for the case where no source config is defined as well as the case where no items are returned for the source. 

I will update the doc string to reflect the new case as well. Thanks for catching that. :)

jgadling: Yeah you're right that it also fills in for missing data, but filling in for *not-found* data is the confusing part to me.

manasaV3: That is for cases such as alignments, if there is no source file available, we should still create a default alignment metadata.

jgadling (approved):

PR #362, opened 2024-11-19
feat: Config for Synaptosome dataset from CZII. 

uermel:
Relates to N/A

## Description
Ingestion config for Synaptosome dataset from CZII.


manasaV3 (approved):

ingestion_tools/dataset_configs/10443.yaml
```diff
@@ -0,0 +1,324 @@
...
+      list_glob: '{run_name}/*.eer'
+
+gains:
+- sources:
+  - source_glob:
+      list_glob: '{gain_file}'
+
+rawtilts:
+- sources:
+  - source_multi_glob:
```
manasaV3, line 172: nit: Is this intended to be a source_multi_glob?

ingestion_tools/dataset_configs/10443.yaml
```diff
@@ -0,0 +1,324 @@
...
+  - literal:
+      value:
+      - '10443'
+
+dataset_keyphotos:
+- sources:
+  - literal:
+      value:
+        snapshot: null
+        thumbnail: null
```
manasaV3, line 105: I believe, we can do away with this.

uermel: That's right, removed now.

ingestion_tools/dataset_configs/10443.yaml
uermel: Indeed not necessary to use that there (and in case of collection metadata). Changed to simple `source_glob`.

PR #367, opened 2024-11-25
fix: Updating ingestion config

manasaV3:
Relates to

## Description
Fixes the alignment and raw tlt sources for dataset 10436.


uermel (approved):
👍

PR #368, opened 2024-11-26
feat: Rescale and threshold input maps during annotation ingestion.

uermel:
Relates to N/A

## Description

Currently, one assumption of the ingestion pipeline for `SemanticSegmentationMask` ingestion is that masks are provided at the same voxel spacing as the tomograms within one dataset's or deposition's config file. Further, it is assumed that segmentation masks are provided as binary or categorical images (i.e. each pixel has an integer value). Upon ingestion, the mask annotation is converted to a binary multiscale pyramid (by selecting the integer value to import using `mask_label`). 

With new datasets, segmentation masks will sometimes be provided at a higher voxel size than the provided tomogram (e.g. in case of #362). In those cases, it would be useful to rescale the segmentation to the tomogram dimensions before ingestion, to ensure the annotation matches the tomogram. 

Another case encountered in a new submission (#363) is that of probability maps. Here, the contributors have provided maps with a continuous value distribution and a default threshold, instead of categorical maps. In the specific case of #363, the maps contain values between 0 and 255 or between 0 and 1, and are to be thresholded at 128 or 0.5, respectively. In case of #363, the input maps are additionally at higher voxel size than the original tomograms.

This PR enables rescaling and thresholding during the ingestion of segmentation masks, and thereby the conversion of input porbability maps to segmentation masks. 

We introduce two new ingestion config parameters for `SemanticSegmentationMask` Annotations, `rescale: bool` and `threshold: float`. When they are not set (default), behavior is the same as previously. When `rescale == True`, the tomogram size in voxels is determined from the parent `VoxelSpacingImporter` and `AlignmentImporter`, and the mask volume is resized to that size. When `threshold is not None`, the (potentially rescaled) image is thresholded at that level.


ingestion_tools/scripts/importers/annotation.py
```diff
@@ -261,19 +274,29 @@ def get_metadata(self, output_prefix: str) -> list[dict[str, Any]]:
 class SegmentationMaskAnnotation(VolumeAnnotationSource):
     shape = "SegmentationMask"  # Don't expose SemanticSegmentationMask to the public portal.
     mask_label: int
+    rescale: bool = False
+    threshold: float | None
```
jgadling, line 278: just a TODO for later that we might want to see if we can/should de-duplicate some of the identical code in SemanticSegmentationMask vs SegmentationMask

ingestion_tools/scripts/tests/s3_import/test_annotations.py
```diff
@@ -1306,6 +1306,73 @@ def test_ingest_instance_point_data(
...
+                    [[0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0]],
+                    [[0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0]],
+                ],
+            },
+        ],
+    },
+    {
+        "case": "SemanticSegmentationMask, small probability map, rescale=True, thresh=0.5, MRC",
+        "source_cfg": {
+            "SemanticSegmentationMask": {
```
jgadling, line 1356: nits: 
- as far as I can tell we're only testing the SemanticSegmentationMask code path. It would be nice to test the SegmentationMask class as well.
- we're testing rescaling without a threshold, but not thresholding without a rescale

jgadling (approved):
Looks pretty sane to me. Would be nice to add the test coverage when you get time.

ingestion_tools/scripts/common/image.py
```diff
@@ -375,12 +386,45 @@ def update_headers(self, mrcfile: MrcFile, header_mapper, voxel_spacing):
...
+
+        if scale_before:
+            data = resize_local_mean(self.volume_reader.get_pyramid_base_data(), self.scale_0_dims, preserve_range=True)
+        else:
+            data = self.volume_reader.get_pyramid_base_data()
+
+        if self.threshold is not None:
+            data = (data >= self.threshold).astype(np.int8)
+
+        data = (data == self.label).astype(np.int8)
```
manasaV3, line 422: Should this check only be done when the threshold value is `None`?

ingestion_tools/scripts/importers/annotation.py
```diff
@@ -282,25 +305,37 @@ def convert(self, output_prefix: str):
...
 
     def __init__(
         self,
         mask_label: int | None = None,
+        rescale: bool = False,
+        threshold: float | None = None,
```
manasaV3, line 323: Not a blocker: Based on the offline discussion, one opportunity for refactor would be to create a new annotation type for cases where the source is a probability map.

manasaV3 (changes requested):

ingestion_tools/scripts/common/image.py
uermel: Yes, the idea is that if threshold is not set, it is not applied.

ingestion_tools/scripts/importers/annotation.py
uermel: Agreed, I'm not sure why there are two classes to be completely honest. They seem to do the same thing.

uermel: Added a note.

manasaV3 (approved):

ingestion_tools/scripts/common/image.py
uermel: Now only comparing against the label when not applying a threshold.

PR #369, opened 2024-11-26
feat: Support for Per Section Alignment Parameters DB Import

kaloster:
Relates to [Add support for perSectionAlignmentParameters to the v2 db ingestion process](https://github.com/chanzuckerberg/cryoet-data-portal/issues/1332)

## Description

Added Per Section Alignment Parameters Importer + Tests

### DIFF
--- a/apiv2/db_import/importer.py
+++ b/apiv2/db_import/importer.py
@@ -45,7 +45,6 @@ def db_import_options(func):
     options.append(click.option("--import-depositions", is_flag=True, default=False))
     options.append(click.option("--import-runs", is_flag=True, default=False))
     options.append(click.option("--import-alignments", is_flag=True, default=False))
-    options.append(click.option("--import-per-section-alignment-parameters", is_flag=True, default=False))
     options.append(click.option("--import-gains", is_flag=True, default=False))
     options.append(click.option("--import-frames", is_flag=True, default=False))
     options.append(click.option("--import-tiltseries", is_flag=True, default=False))
@@ -93,7 +92,6 @@ def load(
     import_depositions: bool,
     import_runs: bool,
     import_alignments: bool,
-    import_per_section_alignment_parameters: bool,
     import_gains: bool,
     import_frames: bool,
     import_tiltseries: bool,
@@ -120,7 +118,6 @@ def load(
         import_depositions,
         import_runs,
         import_alignments,
-        import_per_section_alignment_parameters,
         import_gains,
         import_frames,
         import_tiltseries,
@@ -148,7 +145,6 @@ def load_func(
     import_depositions: bool = False,
     import_runs: bool = False,
     import_alignments: bool = False,
-    import_per_section_alignment_parameters: bool = False,
     import_gains: bool = False,
     import_frames: bool = False,
     import_tiltseries: bool = False,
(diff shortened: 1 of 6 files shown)


apiv2/db_import/tests/test_db_alignment_import.py
```diff
@@ -0,0 +1,132 @@
...
+    sync_db_session.commit()
+    actual = verify_dataset_import(import_per_section_alignment_parameters=True)
+    expected_iter = iter(expected_per_section_alignment_parameters)
+    for run in sorted(actual.runs, key=lambda x: x.name):
+        for alignment in run.alignments:
+            for per_section_alignment_parameters in alignment.per_section_alignment_parameters:
+                expected = next(expected_iter)
+                if "alignment_id" not in expected:
+                    expected["alignment_id"] = alignment.id
+                verify_model(per_section_alignment_parameters, expected)
```
jgadling, line 132: so if I'm reading this test (`expected_per_section_alignment_parameters` only returns a single object) and our test data (there are two items in the `per_section_alignment_parameters` list in the json file) properly, I think we should expect to have *two* per section alignment parameters in our db, but this test is only getting a single row? What's going on here?

test_infra/test_files/30001/RUN1/Alignments/100/alignment_metadata.json
```diff
@@ -76,6 +44,7 @@
   },
   "tilt_offset": 0,
   "x_rotation_offset": 0,
-  "deposition_id": 301,
-  "last_updated_at": 1728675677
+  "deposition_id": 300,
+  "last_updated_at": 1728675677,
+  "tiltseries_id": 501
```
jgadling, line 49: `tiltseries_id` is not a field that exists in the data produced by our S3 ingestion process -- I think we should avoid adding arbitrary fields to our test data.

apiv2/db_import/tests/test_db_alignment_import.py
kaloster: Updated the populate db script accordingly

apiv2/db_import/importer.py
```diff
@@ -45,6 +45,7 @@ def db_import_options(func):
     options.append(click.option("--import-depositions", is_flag=True, default=False))
     options.append(click.option("--import-runs", is_flag=True, default=False))
     options.append(click.option("--import-alignments", is_flag=True, default=False))
+    options.append(click.option("--import-per-section-alignment-parameters", is_flag=True, default=False))
```
manasaV3, line 48: Instead of this flag, should we update this anytime alignments are imported?

jgadling (approved):

apiv2/db_import/tests/test_db_alignment_import.py
```diff
@@ -0,0 +1,146 @@
...
+    actual = verify_dataset_import(import_alignments=True)
+    expected_per_section_alignments_iter = iter(expected_per_section_alignment_parameters)
+    for run in sorted(actual.runs, key=lambda x: x.name):
+        for tiltseries in run.tiltseries:
+            for alignment in tiltseries.alignments:
+                for per_section_alignment in alignment.per_section_alignments:
+                    expected = next(expected_per_section_alignments_iter)
+                    if "alignment_id" not in expected:
+                        expected["alignment_id"] = alignment.id
+                    verify_model(per_section_alignment, expected)
```
jgadling, line 146: It would be nice to add an extra assertion to these tests to check that we called `verify_model` the expected number of times, in order to ensure that we don't wind up skipping any iterations and silently ignoring failure scenarios

PR #370, opened 2024-11-27
fix: Use method type 'simulated' for 10439 and 10441

uermel:
Relates to https://github.com/chanzuckerberg/cryoet-data-portal/issues/1076

## Description
Method type `simulated` was introduced for simulated data, but never updated in the ingestion configs for simulated data. Fixed here.


manasaV3 (approved):
:shipit:

PR #371, opened 2024-12-02
chore: Workaround for py3.12 on osx issue.

jgadling:
zfpy is a sub-sub-sub-dependency of our ingestion scripts, and it does not work properly with Python 3.12 on ARM macs. This workaround only imports the module when it's actually required and allows us to keep running enqueue scripts from our laptops.

(more traceback and info will be added to this ticket later)


ingestion_tools/scripts/importers/visualization_precompute.py
```diff
@@ -90,6 +88,8 @@ def neuroglancer_precompute(self, output_prefix: str, voxel_spacing: float) -> N
         precompute_path = self._get_neuroglancer_precompute_path(annotation_path, output_prefix)
         metadata = self.annotation.metadata
         tmp_path = fs.localwritable(precompute_path)
+        from cryoet_data_portal_neuroglancer.precompute import points
```
manasaV3, line 91: Could you also add a comment for context? :)

kaloster (approved):

PR #372, opened 2024-12-02
fix: db_import updates for alignments

kaloster:
## Description
When running `enqueue_runs.py` with `--import-alignments` flag it will fail for v1 since that flag doesn't exist. This adds this flag to the function definition as a workaround to make sure the job does not fail when importing with the alignments flag.

### DIFF
--- a/ingestion_tools/scripts/db_import.py
+++ b/ingestion_tools/scripts/db_import.py
@@ -27,6 +27,7 @@ def cli():
 
 def db_import_options(func):
     options = []
+    options.append(click.option("--import-alignments", is_flag=True, default=False))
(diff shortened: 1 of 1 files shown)


jgadling (approved):
LGTM

ingestion_tools/scripts/db_import.py
```diff
@@ -27,6 +27,7 @@ def cli():
 
 def db_import_options(func):
     options = []
+    options.append(click.option("--import-alignments", is_flag=True, default=False))
```
manasaV3, line 30: Curious on why we are adding alignment import support to v1 db import?

kaloster: Updated PR description with the reasoning!

PR #383, opened 2024-12-10
chore: fix s3 validation tools

jgadling:
These changes update our S3 validation tools so that they handle our recent directory/schema changes:
- Support multiple tomograms per run
- Support multiple tiltseries per run
- Follow new directory naming conventions

It also adds support for running data validation jobs in AWS via our `enqueue_runs.py` script. When we run jobs in the cloud, the test results are available at the following URL's (just replace the dataset id) depending on whether you're validating staging or prod data:

- https://files.cryoet.staging.si.czi.technology/prod_validation/10168/index.html
- https://files.cryoet.staging.si.czi.technology/staging_validation/10168/index.html

These allure reports *do* include history, for up to the last 20 runs.

This PR is supposed to bring our test coverage back up to parity with where it was at before we broke them by our multiple-tomogram-support migrations. It does *not* attempt to reassess which tests are still aligned to our needs.


ingestion_tools/scripts/data_validation/conftest.py
```diff
@@ -98,6 +98,64 @@ def get_voxel_spacings_set(voxel_spacing_files: List[str]) -> List[str]:
     return list(set(voxel_spacings))
 
 
+def dataset_run_tiltseries_combinations(
```
jgadling, line 101: We had some tests that assumed one tiltseries per run. This method lets us update those tests to support multiple tiltseries.

ingestion_tools/scripts/data_validation/conftest.py
```diff
@@ -98,6 +98,64 @@ def get_voxel_spacings_set(voxel_spacing_files: List[str]) -> List[str]:
...
+        )
+        files = [
+            res_tuple
+            for res_list in files_lists
+            for res_tuple in res_list
+        ]
+
+    return files
+
+def dataset_run_tomogram_combinations(
```
jgadling, line 130: We had some tests that assumed a single tomogram per run. This lets us update them to support multiple tomos per run.

ingestion_tools/scripts/data_validation/fixtures/data.py
```diff
@@ -173,9 +173,9 @@ def tiltseries_metadata(tiltseries_meta_file: str, filesystem: FileSystemApi) ->
 
 
 @pytest.fixture(scope="session")
-def tiltseries_mdoc(tiltseries_mdoc_file: str, filesystem: FileSystemApi) -> pd.DataFrame:
+def frames_mdoc(frames_mdoc_file: str, filesystem: FileSystemApi) -> pd.DataFrame:
```
jgadling, line 176: mdocs are now in the `Frames` directory, rather than associated with tiltseries

ingestion_tools/scripts/data_validation/fixtures/data.py
```diff
@@ -218,17 +218,29 @@ def tomogram_metadata(tomo_meta_file: str, filesystem: FileSystemApi) -> Dict:
     with filesystem.open(tomo_meta_file, "r") as f:
         return json.load(f)
 
+@pytest.fixture(scope="session")
+def all_vs_tomogram_metadata(voxel_dir: str, filesystem: FileSystemApi) -> list[Dict]:
```
jgadling, line 222: Get all tomogram metadatas for this run+voxelspacing

ingestion_tools/scripts/data_validation/fixtures/data.py
```diff
@@ -218,17 +218,29 @@ def tomogram_metadata(tomo_meta_file: str, filesystem: FileSystemApi) -> Dict:
...
+
 
 # ==================================================================================================
 # Neuroglancer fixtures
 # ==================================================================================================
 
 
 @pytest.fixture(scope="session")
-def neuroglancer_config(neuroglancer_config_file: str, filesystem: FileSystemApi) -> Dict:
+def neuroglancer_configs(neuroglancer_config_files: List[str], filesystem: FileSystemApi) -> Dict:
```
jgadling, line 237: Support multiple ng configs per voxelspacing

ingestion_tools/scripts/data_validation/tests/test_alignments.py
```diff
@@ -0,0 +1,90 @@
...
+import allure
+import numpy as np
+import pandas as pd
+import pytest
+from helper_angles import helper_angles_injection_errors
+
+
+@pytest.mark.alignment
+@pytest.mark.parametrize("dataset, run_name, aln_dir", pytest.dataset_run_alignment_combinations, scope="session")
+class TestAlignments:
```
uermel, line 12: We should also validate the AreTomo alignments, which don't necessarily have a `tlt` file. I made [this package](https://github.com/uermel/cryoet-alignment), which we also use for ingestion, so we should also re-use it here. What I would suggest (to not block this PR) is that we consider updating the alignment validations as a separate scope of work, which I can take on, because there are a lot of technical details to be considered.

ingestion_tools/scripts/data_validation/fixtures/data.py
uermel: I think there are things we can validate within the configs themselves, but this is also outside of the scope of this PR I believe.

ingestion_tools/docs/enqueue_runs.md
```diff
@@ -153,6 +153,36 @@ python3 enqueue_runs.py sync --exclude '*' --include 'Annotations/*.json' --s3-p
 | --include-deposition | null    | Look for deposition metadata  with the deposition ids passed here. This helps sync the deposition data.                                                                                            |
 | --no-sync-dataset | False   | Skip syncing datasets. This is useful when we want to only update deposition data                                                                                                                  |
 
+## S3 File validation (`validate` subcommand)
+
+The `enqueue_runs.py validate` command queues up jobs to validate dataset data in s3. It *requires* one or more dataset ID's to be specified. Once the jobs are complete, you can view test results for staging or prod by updating the following URLs with the necessary dataset ID's:
+
+https://files.cryoet.staging.si.czi.technology/prod_validation/10168/index.html
+https://files.cryoet.staging.si.czi.technology/staging_validation/10168/index.html
```
uermel, line 161: This is awesome, but I'm getting an Access Denied message for the staging link. Prod link works.

uermel: image

uermel (changes requested):
This looks great, some changes to the alignment validation, but I think this is maybe something I should take a look at after this is merged?

ingestion_tools/docs/enqueue_runs.md
jgadling: Yep, I haven't run the validation on every dataset .... YET

uermel: Sorry, no rush, I didn't realize that was the reason 😅

ingestion_tools/scripts/enqueue_runs.py
```diff
@@ -598,6 +639,51 @@ def sync(
...
+        input_bucket = "cryoet-data-portal-staging"
+        output_dir = "staging_validation"
+
+    # We always run validation in the staging env.
+    kwargs["environment"] = "staging"
+    handle_common_options(ctx, kwargs)
+
+    # Default to using a lot less memory than the ingestion job.
+    if not ctx.obj.get("memory"):
+        ctx.obj["memory"] = 4000
```
manasaV3, line 674: 🔥

ingestion_tools/scripts/data_validation/tests/test_alignments.py
manasaV3: One suggestion I would offer is to avoid using the same package for generating the alignment values in the validation process. Instead, I recommend explicitly defining it for validation (even though it might be a little redundant), as this would make it easier to identify any potential regressions.

ingestion_tools/scripts/data_validation/tests/test_frame.py
```diff
@@ -52,17 +52,54 @@ def test_extensions(self, frames_files: List[str]):
...
 
-    ### END Self-consistency tests ###
+    ### BEGIN MDOC tests ###
+    @allure.title("Mdoc: tilt angles are within the expected range [-90, 90].")
+    def test_frames_mdoc_range(self, frames_mdoc: pd.DataFrame):
+        assert frames_mdoc["TiltAngle"].min() >= -90
+        assert frames_mdoc["TiltAngle"].max() <= 90
+
+    @allure.title("Mdoc: number of mdoc tilt angles equals number of frames files.")
+    def test_mdoc_frames(self, frames_mdoc: pd.DataFrame, frames_files: List[str]):
```
manasaV3, line 62: I believe this is an existing test, but is it a valid case, when there are lesser number of entries in the mdoc than the actual number of frames?

ingestion_tools/scripts/data_validation/tests/test_frame.py
```diff
@@ -52,17 +52,54 @@ def test_extensions(self, frames_files: List[str]):
...
+        assert len(frames_mdoc) == len(frames_files)
+
+    @allure.title("Mdoc: Every mdoc tilt angle matches a frames file (one-to-one).")
+    def test_mdoc_frame_paths(
```
manasaV3, line 66: Should this be a blocking test or should this be categorized as a warning?

ingestion_tools/scripts/data_validation/tests/test_alignments.py
uermel: I agree, it's just a lot of parsing code to rewrite because the ALN files unfortunately are not simple column-based files. Totally doable though. FWIW, the parser I wrote has a bunch of tests, so maybe we are more interested in making sure those are good and trust the parser?

ingestion_tools/scripts/data_validation/tests/test_alignments.py
```diff
@@ -0,0 +1,90 @@
...
+    def test_tilt_raw_tilt(self, alignment_tilt: pd.DataFrame, alignment_tiltseries_raw_tilt: pd.DataFrame):
+        errors = helper_angles_injection_errors(
+            alignment_tilt["TiltAngle"].to_list(),
+            alignment_tiltseries_raw_tilt["TiltAngle"].to_list(),
+            "tilt file",
+            "raw tilt file",
+        )
+        assert len(errors) == 0, "\n".join(errors)
+
+    @allure.title("Alignment: every tilt angle maps to a mdoc tilt angle.")
```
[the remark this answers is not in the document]
jgadling: Should we remove this test?

ingestion_tools/scripts/data_validation/tests/test_frame.py
uermel: This should be a blocking validation. We should have a frame file for every mdoc section.

ingestion_tools/scripts/data_validation/tests/test_alignments.py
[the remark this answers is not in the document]
uermel: I think this is because before we were not distinguishing `rawtlt` and `tlt` files. It is fine to do this check for `rawtlt`, just not for `tlt`.

uermel (approved):

uermel:
From offline conversation with @jgadling: merge this and deal with fixing alignment validations later.

PR #389, opened 2024-12-20
chore: Cleaning up code and assets that are not used

manasaV3:
Relates to ~

## Description
- Removes script that was used for migrating the data organization, as it no longer used.  
- Removes script for converting mrc to zarr and omezarr as it is no longer used
- Removes script for converting data between json and yaml that is no longer used
- Groups data fetching scripts


jgadling (approved):

jgadling:
fwiw, I'd probably move a bunch of the misc scripts that only get used once in awhile into a `utils` directory instead of a `data_importers` directory, basically try to mark them as something we rarely need.

PR #394, opened 2025-01-03
feat: support filtering on related aggregates

jgadling:
## Description

Relates to / fixes [#1417](https://github.com/chanzuckerberg/cryoet-data-portal/issues/1417) and implements filtering on aggregates almost identically to the way that Hasura supports aggregate filters.

Screenshot 2025-01-03 at 3 54 06 PM

### DIFF
--- a/apiv2/platformics/graphql_api/core/query_builder.py
+++ b/apiv2/platformics/graphql_api/core/query_builder.py
@@ -83,7 +83,10 @@ def convert_where_clauses_to_sql(
     # Create a dictionary with the keys as the related field/field names
     # The values are dict of {order_by: {"field": ..., "index": ...}, where: {...}, group_by: [...]}
     all_joins = defaultdict(dict)  # type: ignore
+    # Keep track of the joins we need to execute in order to make filtering by relationships work.
     where_joins = defaultdict(dict)  # type: ignore
+    # Keep track of the joins we need to execute to handle filtering on related aggregates.
+    aggregate_joins = defaultdict(dict)  # type: ignore
     for item in order_by:
         for col, v in item["field"].items():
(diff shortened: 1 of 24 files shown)


apiv2/platformics/graphql_api/core/query_builder.py
```diff
@@ -130,6 +138,39 @@ def convert_where_clauses_to_sql(
         for local, remote in relationship.local_remote_pairs:
             subquery = subquery.filter((getattr(sa_model, local.key) == getattr(query_alias, remote.key)))
         query = query.where(subquery.exists())
+    # Handle filtering on aggregates
+    for aggregate_field, aggregate_info in aggregate_joins.items():
+        relationship = mapper.relationships[aggregate_field]  # type: ignore
+        related_cls = relationship.mapper.entity
```
manasaV3, line 144: Do we need a validation of the users authorization to query or aggregate on this class?

jgadling: We 100% do! All queries and subqueries and joins *must include* authorization filters!

In this case, those filters are applied when we call `get_aggregate_db_query` a few lines below this point which (eventually) calls `get_resource_query` which constructs a query for a given resource type with authorization policies already applied. 

From that point on, we're working with a SQLAlchemy query object that represents something like `select count(*) from some_table where user_id = 11`. All the code in `query_builder.py` just adds additional filtering/ordering/etc criteria to these queries, without removing the filters that enforce authz policy.

the `relationship` and `related_class` variables here just help us identify which relationship we're working with here, so we know how to join an outer query to a subquery.

manasaV3 (approved):

PR #397, opened 2025-01-07
fix: support __typename fields in gql queries

jgadling:
Adds support for including `__typename` and other graphql-reserved keywords in GraphQL queries. Our API was making naive/incorrect assumptions about query field ordering and also assuming that only application-specific fields would be present in queries. This updates our application code to filter out any GraphQL reserved fields (such as `__typename`) before interpreting queries. 

Some GraphQL clients (like Apollo) include `__typename` fields in every level of a GraphQL query to help with type interpretation. We've had to disable this functionality in our frontend to work around the issue in our API. We should be able to turn it back on now.

Screenshot 2025-01-08 at 11 07 17 AM

### DIFF
--- a/apiv2/platformics/graphql_api/core/strawberry_helpers.py
+++ b/apiv2/platformics/graphql_api/core/strawberry_helpers.py
@@ -1,0 +1,35 @@
+from strawberry.types.nodes import SelectedField
+from platformics.graphql_api.core.errors import PlatformicsError
+from typing import Tuple
+
+def filter_meta_fields(selections: list[SelectedField]) -> list[SelectedField]:
+    return [item for item in selections if not item.name.startswith("__")]
+
+def get_field_by_name(selections: list[SelectedField], item_name: str) -> SelectedField:
+    for item in selections:
+        if item.name == item_name:
+            return item
+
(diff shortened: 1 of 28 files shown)


apiv2/platformics/graphql_api/core/strawberry_helpers.py
```diff
@@ -0,0 +1,41 @@
+from typing import Tuple
+
+from strawberry.types.nodes import SelectedField
+
+from platformics.graphql_api.core.errors import PlatformicsError
```
jgadling, line 5: This file basically adds some helper functions that filter out graphql reserved fields (they start with `__`) from a list of fields in a query.

apiv2/tests/helpers.py
```diff
@@ -0,0 +1,35 @@
+import operator
+
+
+# Helper for checking deep equality of python dict/list objects
+# Source: https://gist.github.com/samuraisam/901117/521ed1ff8937cb43d7fcdbc1a6f6d0ed2c723bae
+def deep_eq(_v1, _v2):
```
jgadling, line 6: Writing piecemeal tests for graphql response structures was getting tedious. I added this so we can validate literal responses.

bchu1 (approved):
Thank you!

PR #399, opened 2025-01-08
fix: Updating schema toml for poetry build update

manasaV3:
Relates to ~

## Description
Fixes the issue with schema's build.


uermel (approved):
Looks good to me, this fixes the GH action.

PR #401, opened 2025-01-08
feat: Add file size to schema

kaloster:
Relates to [Capture file size in the database](https://github.com/orgs/chanzuckerberg/projects/47/views/1?pane=issue&itemId=52284203&issue=chanzuckerberg%7Ccryoet-data-portal%7C444)

## Description
- [x] Update db schema
- [x] Update import scripts
- [x] Update Integer to BigInteger
- [x] Add pagination
- [x] Update tests
- [ ] Run import scripts for all datasets
- [ ] Update client
- [ ] Update documentation


apiv2/schema/schema.yaml
```diff
@@ -841,6 +841,17 @@ classes:
...
+        name: file_size
+        description: Size of the file in bytes
+        from_schema: cdp-dataset-config
+        exact_mappings:
+          - api_file_size
+        range: integer
+        minimum_value: 0
+        unit:
+          symbol: MB
+          descriptive_name: megabytes
```
jgadling, line 854: It's typical to return file sizes in bytes and then convert filesizes to (mb/gb/etc when those sizes are displayed. I'd prefer if we stored sizes in bytes, but if we specifically want to store them in megabytes for some reason, can we change the field names to `file_size_mb` or something to ensure that nobody tries to read them as bytes?

apiv2/db_import/importers/base_importer.py
```diff
@@ -79,6 +79,23 @@ def get_s3_url(self, *input_path: tuple[str]) -> str:
...
         return os.path.join(self.config.s3_prefix, input_path)
 
+    def get_file_size(self, *input_path: tuple[str]) -> str:
+        input_path = os.path.join(*input_path)
+        if input_path.startswith(self.config.bucket_name):
+            input_path = input_path[len(self.config.bucket_name) + 1 :]
+
+        total_size = 0
+        try:
+            response = self.config.s3_client.list_objects_v2(Bucket=self.config.bucket_name, Prefix=input_path)
```
jgadling, line 89: Do we need to do any pagination here? I think the S3 api only returns 1000 files per request, and I suspect (??) that for some datasets we have more than 1000 files.

apiv2/db_import/importers/base_importer.py
```diff
@@ -79,6 +79,23 @@ def get_s3_url(self, *input_path: tuple[str]) -> str:
...
+            if 'Contents' in response:
+                for obj in response['Contents']:
+                    total_size += obj['Size']
+
+            total_size_mb = round(total_size / (1024 * 1024))
+            return total_size_mb if total_size_mb > 1 else None
```
jgadling, line 95: Yeah, let's not convert to mb here, the FE can convert bytes to something human-readable.

jgadling (approved):

PR #402, opened 2025-01-08
fix: Don't replace spaces with underscores for reconstruction software.

jgadling:
We were erroneously replacing spaces with underscores for the reconstruction software field when importing metadata to the db. This fixes the issue and updates the tests to reflect the change.

Relates to


manasaV3 (approved):
🚢

apiv2/db_import/importers/tomogram.py
```diff
@@ -62,7 +62,7 @@ def load_computed_fields(self):
                 self.input_data.get("fiducial_alignment_status", False),
             ),
             "reconstruction_method": self.normalize_to_unknown_str(self.input_data.get("reconstruction_method")),
```
manasaV3, line 64: Should we doing the same for the `reconstruction_method`?

PR #403, opened 2025-01-09
fix: Fix an example command for enqueue_runs

uermel:
Relates to N/A

## Description
Minor fix for the enqueue runs examples.


jgadling (approved):

uermel:
@jgadling Seems one of the db import tests is failing?

PR #407, opened 2025-01-10
fix: Include threshold when checking mask annotation validity

uermel:
Relates to N/A

## Description
When checking if a SegmentationMask annotation is valid the threshold (if it exists) needs to be passed to the MaskConverter. When a threshold is applied, the mask is checked for any value > threshold instead of simply checking for the presence of the mask label. This PR fixes that by adding the threshold as an optional argument to the validity check.

Required for correctly ingesting the Ais deposition. (#363)


jgadling (approved):
LGTM

PR #408, opened 2025-01-10
chore: Updating long execution name generation

manasaV3:
Relates to ~

Addresses the issue mentioned above.


uermel (approved):
Looks good to me, and thanks for fixing the DB test as well!

PR #417, opened 2025-01-17
revert: Add file size to schema (#401)

kaloster:
This reverts commit fc98398745797c2e7fb9db32086b88194529c32b.

Relates to https://github.com/chanzuckerberg/cryoet-data-portal-backend/pull/401


docker-compose.yml
```diff
@@ -6,7 +6,7 @@ services:
     volumes:
       - ./test_infra/test_files:/usr/share/nginx/html
   motoserver:
-    image: motoserver/moto:5.0.26
```
jgadling, line 9: NOTE -- we may want to re-upgrade moto in a future PR!

jgadling (approved):

docker-compose.yml
kaloster: noted!

PR #421, opened 2025-01-22
feat: add content type for s3 ingestion

kaloster:
Relates to https://github.com/chanzuckerberg/cryoet-data-portal/issues/1390

## Description

Added `setxattr` to S3 ingestion for `ContentType` attribute

### DIFF
--- a/ingestion_tools/scripts/common/fs.py
+++ b/ingestion_tools/scripts/common/fs.py
@@ -1,5 +1,6 @@
 import contextlib
 import glob
+import mimetypes
 import os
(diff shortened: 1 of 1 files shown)


manasaV3 (approved):

manasaV3 (commented):
Sorry, I found this just now, https://s3fs.readthedocs.io/en/latest/api.html#s3fs.core.S3FileSystem.put_file supports kwargs, we could pass `ContentType` to that optionally. That would make it a single call.

PR #422, opened 2025-01-23
revert: content type

kaloster:
This reverts - https://github.com/chanzuckerberg/cryoet-data-portal-backend/pull/421


manasaV3 (approved):
This needed to be reverted, because it would set the content type for images to also `application/octet-stream`.

PR #425, opened 2025-01-23
fix: add content type to metadata file

kaloster:
Relates to https://github.com/chanzuckerberg/cryoet-data-portal/issues/1390

## Description
 This one makes sure that `ContentType` is set only for metadata json files

Tested [here](https://us-west-2.console.aws.amazon.com/s3/object/cryoetportal-output-test?region=us-west-2&bucketType=general&prefix=10000/dataset_metadata.json) and [here](https://us-west-2.console.aws.amazon.com/s3/object/cryoetportal-output-test?region=us-west-2&bucketType=general&prefix=10002/dataset_metadata.json).

### DIFF
--- a/ingestion_tools/scripts/common/fs.py
+++ b/ingestion_tools/scripts/common/fs.py
@@ -90,6 +90,10 @@ class S3Filesystem(FileSystemApi):
         return self.s3fs.glob(*args)
 
     def open(self, path: str, mode: str, **kwargs) -> TextIOBase:
+        _, file_extension = os.path.splitext(path)
+        content_type = "application/json" if file_extension == ".json" else None
+        if content_type:
+            kwargs['ContentType'] = content_type
         return self.s3fs.open(path, mode, **kwargs)
 
     def localreadable(self, path: str) -> str:
--- a/ingestion_tools/scripts/common/fs.py
+++ b/ingestion_tools/scripts/common/fs.py
@@ -90,10 +90,11 @@ class S3Filesystem(FileSystemApi):
         return self.s3fs.glob(*args)
 
     def open(self, path: str, mode: str, **kwargs) -> TextIOBase:
-        _, file_extension = os.path.splitext(path)
-        content_type = "application/json" if file_extension == ".json" else None
-        if content_type:
-            kwargs['ContentType'] = content_type
+        if mode == "w":
+            _, file_extension = os.path.splitext(path)
+            content_type = "application/json" if file_extension == ".json" else None
+            if content_type:
+                kwargs['ContentType'] = content_type
         return self.s3fs.open(path, mode, **kwargs)
 
     def localreadable(self, path: str) -> str:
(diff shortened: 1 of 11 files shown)


ingestion_tools/scripts/common/fs.py
```diff
@@ -90,6 +90,10 @@ def glob(self, *args: list[str]) -> list[str]:
         return self.s3fs.glob(*args)
 
     def open(self, path: str, mode: str, **kwargs) -> TextIOBase:
+        _, file_extension = os.path.splitext(path)
+        content_type = "application/json" if file_extension == ".json" else None
+        if content_type:
+            kwargs['ContentType'] = content_type
```
jgadling, line 96: Hrm, trying to set headers on a file that we could be *reading* (if `mode` includes `r` for example) sounds weird. Does this cause any problems for json file reading?

kaloster: Added condition for `w` mode only.

ingestion_tools/scripts/tests/s3_import/test_collection_metadata.py
```diff
@@ -21,9 +21,9 @@ def test_collection_metadata_import(s3_fs: FileSystemApi, test_output_bucket: st
     source_filename = os.path.basename(collection_metadata[0].path)
     assert source_filename in actual_files
 
-    actual = get_data_from_s3(s3_client, test_output_bucket, os.path.join(prefix, source_filename)).readlines()
+    actual = get_data_from_s3(s3_client, test_output_bucket, os.path.join(prefix, source_filename))["Body"].readlines()
     source_file_path = "/".join(collection_metadata[0].path.split("/")[1:])
-    expected = get_data_from_s3(s3_client, "test-public-bucket", source_file_path).readlines()
+    expected = get_data_from_s3(s3_client, "test-public-bucket", source_file_path)["Body"].readlines()
```
jgadling, line 26: Uhh, why did the interface change?

kaloster: Because of an update to `get_data_from_s3()` to get `ContentType` as well in a single call and make the assertion in `validate_metadata()`

kaloster: Added a separate interface for raw data

jgadling (approved):
Can you add a test to ensure that non-json files are written *without* a content-type: json header? 

Ok to merge after that, thanks for the fixes!

ingestion_tools/scripts/common/metadata.py
```diff
@@ -34,7 +34,8 @@ class MergedMetadata(BaseMetadata):
     def write_metadata(self, filename: str, merge_data: dict[str, Any]) -> None:
         metadata = deep_merge(self.metadata, merge_data)
         metadata = self.add_defaults(metadata)
-        with self.fs.open(filename, "w") as fh:
+        kwargs = {"ContentType": "application/json"}
+        with self.fs.open(filename, "w", **kwargs) as fh:
```
manasaV3, line 38: Should this be conditionally added to `BaseMetadata`, where `is_json==True`?

kaloster: Added

PR #428, opened 2025-02-01
feat: Migrating existing configs to include frame & ctf metadata

manasaV3:
Relates to https://github.com/chanzuckerberg/cryoet-data-portal/issues/1647

Depends on: https://github.com/chanzuckerberg/cryoet-data-portal-backend/pull/427

## Description
- Manually migrated the ctf entity
- Updates the generated rawtlt and mdoc files as fallback
- Adds migration to support frame_dose_rate 

### Notes:

1. There is a dependency on file `run_dose_rate_map.tsv` to exist in `https://github.com/czimaginginstitute/czii-data-portal-processing/tree/main/src/data_portal_processing/jensendb` containing a mapping of all run names to frame dose rating.

2. The failure of the ingestion config validation is expected due invalid stub placeholder, as a remainder to update these frame dose rate values. Do not merge until all the errors have been addressed.


jgadling (approved):

ingestion_tools/scripts/data_validation/shared/helper/angles_helper.py
```diff
@@ -13,7 +12,7 @@ def helper_angles_injection_errors(
     remaining_angles = codomain_angles.copy()
     for domain_angle in domain_angles:
         found_match = False
-        for codomain_angle in codomain_angles:
+        for codomain_angle in remaining_angles:
```
uermel, line 15: We need to check for presence in the remaining angles, since we have cases where two angles are very close (below the tolerance value). If we keep checking against the codomain angles this leads to the incorrect one being identified (which should have already been removed).

ingestion_tools/scripts/data_validation/standardized/fixtures/path.py
```diff
@@ -84,7 +84,7 @@ def frames_files(frames_dir: str, filesystem: FileSystemApi) -> List[str]:
     """[Dataset]/[ExperimentRun]/Frames/*"""
     files = filesystem.glob(f"{frames_dir}/*")
     # Exclude mdoc files, add s3 prefix
-    refined_files = ["s3://" + file for file in files if ".mdoc" not in file]
+    refined_files = ["s3://" + file for file in files if ".mdoc" not in file and ".json" not in file]
```
uermel, line 87: Frames metadata exist now and need to be excluded from count.

ingestion_tools/scripts/data_validation/standardized/tests/test_tiltseries.py
```diff
@@ -111,18 +117,21 @@ def test_mdoc_tiltseries_metadata(self, tiltseries_metadata: dict, mdoc_data: pd
...
             "tiltseries metadata tilt_range",
+            "mdoc file",
         )
         assert len(errors) == 0, (
-                "\n".join(errors)
-                + f"\nRange: {tiltseries_metadata['tilt_range']['min']} to {tiltseries_metadata['tilt_range']['max']}, "
-                  f"with step {tiltseries_metadata['tilt_step']}"
+            "\n".join(errors)
+            + f"\nRange: {tiltseries_metadata['tilt_range']['min']} to {tiltseries_metadata['tilt_range']['max']}, "
+            f"with step {tiltseries_metadata['tilt_step']}"
```
uermel, line 134: Needed to invert the logic here, see description of the test. `helper_angles_injection_errors ` checks if domain angles are also in the codomain. Since the mdoc may contain more angles than the tilt series, tilt series metadata range needs to be domain, mdoc needs to be codomain.

ingestion_tools/scripts/gjensen_config.py
```diff
@@ -285,11 +288,17 @@ def to_template_by_run(templates, run_data_map, prefix: str, path: list[str]) ->
     return template_metadata
 
 
-def to_tiltseries(data: dict[str, Any]) -> dict[str, Any]:
+def to_tiltseries(
+    per_run_float_mapping: dict[str, dict[str, float]],
+    per_run_string_mapping: dict[str, dict[str, str]],
```
uermel, line 293: Adding overrides for float and string values here.

ingestion_tools/scripts/gjensen_config.py
```diff
@@ -303,11 +312,21 @@ def to_tiltseries(data: dict[str, Any]) -> dict[str, Any]:
...
     tilt_series["tilt_range"] = {
-        "min": tilt_series.pop("tilt_range_min"),
-        "max": tilt_series.pop("tilt_range_max"),
+        "min": per_run_float_mapping["tilt_series_min_angle"][run_name],
+        "max": per_run_float_mapping["tilt_series_max_angle"][run_name],
     }
+    tilt_series["tilt_step"] = per_run_float_mapping["tilt_series_tilt_step"][run_name]
+    tilt_series["tilting_scheme"] = per_run_string_mapping["tilt_series_tilting_scheme"][run_name]
+    tilt_series["tilt_axis"] = per_run_float_mapping["tilt_series_tilt_axis"][run_name]
+    tilt_series["total_flux"] = per_run_float_mapping["tilt_series_total_flux"][run_name]
```
uermel, line 321: Since these values need to be overridden for at least some runs, they are all taken from the override map.

ingestion_tools/scripts/gjensen_config.py
```diff
@@ -569,6 +588,80 @@ def get_cross_reference_mapping(input_dir: str) -> dict[int, dict[str, str]]:
...
+    runs = data["runs"]
+    runs_out = []
+    for entry in runs:
+        if run_include_mapping.get(entry["run_name"], True):
+            runs_out.append(entry)
+        else:
+            print(f"Excluding run {entry['run_name']}")
+
+    data["runs"] = runs_out
+    return data
```
uermel, line 662: Simple boolean flag for each run on whether to include in the config run-regex or not.

ingestion_tools/scripts/gjensen_config.py
```diff
@@ -569,6 +588,80 @@ def get_cross_reference_mapping(input_dir: str) -> dict[int, dict[str, str]]:
...
+    """
+    with open(os.path.join(input_dir, "per_run_string_param_map.tsv")) as csvfile:
+        reader = csv.DictReader(csvfile, delimiter="\t")
+        params = reader.fieldnames
+        ret = {param_name: {} for param_name in params if param_name != "run_name"}
+
+        for row in reader:
+            for param_name in ret:
+                ret[param_name][row["run_name"]] = row[param_name]
+        return ret
```
uermel, line 626: Float and string overrides to use during config generation.

ingestion_tools/scripts/gjensen_config.py
uermel, line 662: Boolean flags for each run are used to exclude them from the CZII-json data before any further processing.

ingestion_tools/scripts/gjensen_config.py
```diff
@@ -581,6 +674,48 @@ def exclude_runs_parent_filter(entities: list, runs_to_exclude: list[str]) -> No
...
+        elif len(values) == 1:
+            ret[param_name] = next(iter(values.keys()))
+        else:
+            key = f"float {{{param_name}}}"
+            for param_value, runs in distinct_values[param_name].items():
+                for run_name in runs:
+                    run_data_map[run_name][param_name] = param_value
+            ret[param_name] = key
+
+    return ret, run_data_map
```
uermel, line 716: Converting from per-run-map to single value or mapped-value entry.

ingestion_tools/scripts/gjensen_config.py
```diff
@@ -663,10 +813,21 @@ def create(ctx, input_dir: str, output_dir: str) -> None:
             f"^{run['run_name']}$" for run in val.get("runs") if not run.get("tilt_series", {}).get("tilt_series_path")
         ]:
             exclude_runs_parent_filter(updated_dataset_config.get("tiltseries", []), runs_without_tilt)
+            exclude_runs_parent_filter(updated_dataset_config.get("frames", []), runs_without_tilt)
+            exclude_runs_parent_filter(updated_dataset_config.get("ctfs", []), runs_without_tilt)
+            exclude_runs_parent_filter(updated_dataset_config.get("rawtlts", []), runs_without_tilt)
+            exclude_runs_parent_filter(updated_dataset_config.get("collection_metadata", []), runs_without_tilt)
```
uermel, line 819: When a run has no tilt series, all of these should have the same exclude filters.

ingestion_tools/scripts/gjensen_config.py
```diff
@@ -663,10 +813,21 @@ def create(ctx, input_dir: str, output_dir: str) -> None:
...
         if runs_without_tomogram := [f"^{run['run_name']}$" for run in val.get("runs") if not run.get("tomograms")]:
             exclude_runs_parent_filter(updated_dataset_config.get("voxel_spacings", []), runs_without_tomogram)
 
+        # If there are no tiltseries, remove frames, rawtlts, ctfs, and collection_metadata
+        if not updated_dataset_config.get("tiltseries"):
+            updated_dataset_config.pop("frames", None)
+            updated_dataset_config.pop("rawtlts", None)
+            updated_dataset_config.pop("ctfs", None)
+            updated_dataset_config.pop("collection_metadata", None)
+
```
uermel, line 830: When a dataset has no tilt series at all, all of these should not be present.

ingestion_tools/scripts/schema_migration/migrate_v1_2_0.py
```diff
@@ -0,0 +1,34 @@
...
+            },
+        ]
+
+    if "frames" in config:
+        for entry in config["frames"]:
+            if "metadata" not in entry:
+                entry["metadata"] = {
+                    "dose_rate": frame_dose_rate,
+                    "is_gain_corrected": "gain" in config,
+                }
```
uermel, line 23: If no tiltseries are present for this dataset, a frames block should not exist. But we need to include an additional check on L17 to only do updates when the block does exist.

ingestion_tools/docs/jensen_config_generation.md
```diff
@@ -0,0 +1,91 @@
+# Generating ingestion config files for Jensen Datasets
```
manasaV3, line 1: 💜

ingestion_tools/scripts/data_validation/standardized/fixtures/path.py
```diff
@@ -84,7 +84,7 @@ def frames_files(frames_dir: str, filesystem: FileSystemApi) -> List[str]:
     """[Dataset]/[ExperimentRun]/Frames/*"""
     files = filesystem.glob(f"{frames_dir}/*")
     # Exclude mdoc files, add s3 prefix
```
manasaV3, line 86:
```suggestion
    # Exclude mdoc and frames_metadata.json files, add s3 prefix
```

ingestion_tools/scripts/gjensen_config.py
```diff
@@ -581,6 +674,48 @@ def exclude_runs_parent_filter(entities: list, runs_to_exclude: list[str]) -> No
             source["parent_filters"]["exclude"]["run"].extend(runs_to_exclude)
 
 
+def handle_per_run_param_maps(
```
manasaV3, line 677: If I am understanding this method correctly, it handles for distinct values in the field, and adds it to `run_data_map`. We later sets that value to the relevant config field while processing the tiltseries. 

What are your thoughts on overriding those values in the tiltseries object of the entry directly here instead? I think it might simplify things a little, and also the csv files that gets generated would not include columns that don't get referenced in the config. For example: the [run_data_map/10014.csv](https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/480ce46a8f48e6a504556237f04e9b4a3c4f3487/ingestion_tools/dataset_configs/gjensen/run_data_map/10014.csv) now includes columns for  `tilt_series_max_angle, tilt_series_min_angle, tilt_series_quality_score` and `ts-tilt_range-max, ts-tilt_range-min, ts-tilt_series_quality`.

While that would work for overrides made to entities such as tiltseries, tomogram. This kind of processing would still be needed for `frames_dose_rate`, unless we refactor how frame_dose_rate is handled now.

ingestion_tools/scripts/gjensen_config.py
manasaV3: Could we a have a usecase, where a frame block was added for a manually generated config but they don't submit a tiltseries with it?

uermel: This is possible, but is it relevant for jensen config generation?

ingestion_tools/scripts/gjensen_config.py
uermel: So the suggestion would be to: 

1. Add dose rate to `float_fields`
2. Override all tilt series values with actual numbers, do not check for distinctness here
3. Let `to_template_by_run` handle the distinctness check

tbh, this function is just a generalization of what was being done for dose rate. I didn't realize there was an extra distinctness check that is called after `to_tiltseries`, and I didn't realize these values were added to the csv twice for that reason. We can do away with the distinctness check ahead of `to_tiltseries` I think.

ingestion_tools/scripts/gjensen_config.py
```diff
@@ -620,6 +764,11 @@ def create(ctx, input_dir: str, output_dir: str) -> None:
         authors = to_author(val.get("dataset")["authors"], deposition_entity.get("metadata", {}).get("authors"))
 
         run_data_map = defaultdict(dict)
+        ds_per_run_mapping, run_data_map = handle_per_run_param_maps(
+            val.get("runs"),
+            run_data_map,
+            per_run_float_mapping,
```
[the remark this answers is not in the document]
uermel: Good point ... but maybe we just remove this altogether, as discussed below?

ingestion_tools/scripts/gjensen_config.py
manasaV3: Sorry, this response was meant to be added to [the comment below](https://github.com/chanzuckerberg/cryoet-data-portal-backend/pull/428#discussion_r2045734032). 😅

ingestion_tools/scripts/gjensen_config.py
manasaV3: iirc, we currently don't do `to_template_by_run` for frame dose_rate. So, we should either update how frames entity is being generated or retain the above method for just frames_dose_rate.

ingestion_tools/scripts/gjensen_config.py
uermel: But for other configs we never need to run this migration again, or am I misunderstanding how these migrations work? 

Do we always apply all migration scripts when we do a new one?

Bento007: > Do we always apply all migration scripts when we do a new one?

No, we only migrate configs as needed if they are not the latest, and we only apply the migration step needed to get it to the latest.

manasaV3: The only config that should ideally be impacted by this is [10006.yaml](https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/main/ingestion_tools/dataset_configs/10006.yaml#L203-L204). But, since it has a tiltseries block defined it would still generate the frame block for it.

ingestion_tools/scripts/gjensen_config.py
```diff
@@ -238,6 +238,7 @@ def to_standardization_config(
     "ts-tilt_range-min",
     "ts-tilt_range-max",
     "ts-total_flux",
+    "frame_dose_rate",
```
manasaV3, line 241: I believe, the addition here will not have the desired impact as we haven't changed how we process for the frames entity.

ingestion_tools/scripts/gjensen_config.py
```diff
@@ -707,12 +706,9 @@ def handle_per_run_param_maps(
             ret[param_name] = next(iter(values.keys()))
         else:
             key = f"float {{{param_name}}}"
-            for param_value, runs in distinct_values[param_name].items():
-                for run_name in runs:
-                    run_data_map[run_name][param_name] = param_value
             ret[param_name] = key
 
-    return ret, run_data_map
```
manasaV3: As we are removing the addition to run_data_map, the  case where there are different values for `frame_dose_rate` for runs in a dataset is not handled correctly.

Bento007 (approved):

PR #429, opened 2025-02-04
feat: Adds metadata generation for Frames and PerSectionParameter

manasaV3:
Relates to https://github.com/chanzuckerberg/cryoet-data-portal/issues/1055, https://github.com/chanzuckerberg/cryoet-data-portal/issues/1056 

## Description
- Moves the ingestion of CollectionMetadata to be before Frames
- Updates Frames Importer to iterate through all the frames of a run together instead each frame being imported separately
- Adds frame_metadata.json
- Adds CTF Importer
- Adds per_section_parameter to `tiltseries_metadata.json`

### TODO:
- [x] Updates the source validation's use of frame importer


ingestion_tools/scripts/importers/base_importer.py
```diff
@@ -301,12 +302,14 @@ class BaseFileImporter(BaseImporter):
     This class extends the BaseImporter to handle the import of files. This is suited for case where no transformations
     are required, and source files can be copied to the destination.
     """
+    def get_destination_path(self) -> str:
```
jgadling, line 305: just a reminder that we probably want to update the rest of our importers to use this when we get a chance!

jgadling (approved):

ingestion_tools/scripts/importers/tiltseries.py
```diff
@@ -92,6 +97,10 @@ def import_metadata(self) -> None:
         merge_data["frames_count"] = self.get_frames_count()
         base_metadata = self.get_base_metadata()
         merge_data["pixel_spacing"] = self.get_pixel_spacing()
+        psp_generator = PerSectionParameterGenerator(self.config, {**self.parents, "tiltseries": self})
```
jgadling, line 100: Just a headsup that if PerSectionParameterGenerator was a class property and called from that property instead of referenced directly here, it could be overridden in tests easily. It looks like you don't strictly *need* to override this in your tests, but it's another tool you could use to inject dependencies better.

PR #430, opened 2025-02-05
fix: Deprecate API V1

kaloster:
Relates to https://github.com/chanzuckerberg/cryoet-data-portal/issues/1449

## Description

PR to deprecate the V1 API from the CryoET Data Portal Backend

- [x] Update github workflows
- [x] Update Docker 
- [x] Update Makefiles
- [x] Remove V1 code
- [x] Update `enqueue_runs` references
- [x] Verify V1 is fully deprecated on frontend
- [x] Coordinate timing with @junxini 
- [x] Delete Happy dev and staging stack
- [x] Delete Happy prod stack
- [x] Remove `.happy` config if nothing breaks
- [ ] Merge it.


.github/workflows/ingestor-test.yml
```diff
@@ -21,5 +21,4 @@ jobs:
         run: |
           make ingestor-init
           make ingestor-test-db-init
```
jgadling, line 23: I think we can take down the `ingestor-test-db-init` stuff as well.

Makefile
```diff
@@ -20,12 +20,6 @@ ingestor-init:
 	$(docker_compose) --profile ingestor up -d
 	cd ./test_infra/; ./seed_moto.sh
 
-.PHONY: api-init
-api-init:
-	docker compose --profile apiv1 up -d
-	docker compose cp test_infra/sql db:/tmp/sql
-	docker compose exec db sh -c 'cat /tmp/sql/seed_db_data.sql | psql postgres://postgres:postgres@127.0.0.1:5432/cryoet'
```
jgadling, line 27: Right now the docker-compose.yml file has the db come up with a `cryoet` schema by default for the v1 api (that we don't need anymore), and then there's a step in this Makefile that creates the `cryoetv2` schema for apiv2. I think we can update the docker-compose.yml file to just start up with a `cryoetv2` database by default and remove the steps that create the second DB from this Makefile

Makefile
```diff
@@ -61,15 +51,8 @@ push-local-ingestor-build:
 	aws_region=$$(aws configure get region); \
 	account_id=$$(aws sts get-caller-identity | jq -r ".Account"); \
 	ecr_repo=$$account_id.dkr.ecr.$$aws_region.amazonaws.com; \
-	$(MAKE) push-ingestor-build ecr_repo=$$ecr_repo/cryoet-staging tag=$(tag) aws_region=$$aws_region; \
```
jgadling, line 64: We still need to build&push the ingestor image because this is what does our S3 ingestion

ingestion_tools/scripts/enqueue_runs.py
```diff
@@ -12,11 +12,11 @@
 from boto3 import Session
 from botocore import UNSIGNED
 from botocore.config import Config
-from db_import import db_import_options
+from db_import.common.config import DBImportConfig
+from db_import.importer import db_import_options
+from db_import.importers.dataset import DatasetDBImporter
+from db_import.importers.deposition import DepositionDBImporter
```
jgadling, line 18: These files all just went away -- I think these imports need to be removed

kaloster: These are needed for the script to run - imports were updated to V2 and no longer point to `importers.db` which was removed

ingestion_tools/scripts/enqueue_runs.py
```diff
@@ -172,8 +172,9 @@ def get_datasets(
     filter_datasets = [re.compile(pattern) for pattern in filter_datasets]
     exclude_datasets = [re.compile(pattern) for pattern in exclude_dataset]
     s3_config = Config(signature_version=UNSIGNED) if anonymous else None
-    s3_client = boto3.client("s3", config=s3_config)
-    config = DBImportConfig(s3_client, None, s3_bucket, https_prefix)
+    session = Session()
+    s3_client = session.client("s3", config=s3_config)
+    config = DBImportConfig(s3_client, None, s3_bucket, https_prefix, session)
```
jgadling, line 177: the `session` argument here is supposed to be a DB session, not an aws session - can we pass in None or refactor the code so that we can instantiate this without a db session?

Makefile
```diff
@@ -44,10 +37,6 @@ ingestor-test-db-init:
 	docker compose exec db sh -c 'cat /tmp/sql/schema.sql | psql postgres://postgres:postgres@127.0.0.1:5432/cryoet'
 	docker compose exec db sh -c 'cat /tmp/sql/seed_db_enum.sql | psql postgres://postgres:postgres@127.0.0.1:5432/cryoet'
```
jgadling, line 38: I'm pretty sure we don't need this anymore. We also have some documentation files that refer to this that may need to be updated

ingestion_tools/scripts/enqueue_runs.py
```diff
@@ -172,7 +172,8 @@ def get_datasets(
     filter_datasets = [re.compile(pattern) for pattern in filter_datasets]
     exclude_datasets = [re.compile(pattern) for pattern in exclude_dataset]
     s3_config = Config(signature_version=UNSIGNED) if anonymous else None
-    s3_client = boto3.client("s3", config=s3_config)
+    session = Session()
+    s3_client = session.client("s3", config=s3_config)
```
jgadling, line 176: This doesn't need to change anymore!

jgadling (approved):

PR #438, opened 2025-02-13
feat: Ingestion config for synaptic vesicle annotations.

uermel:
Relates to N/A

## Description
Semantic and Instance segmentations of synaptic vesicles in dataset 10443 created using [SynapseNet](https://www.biorxiv.org/content/10.1101/2024.12.02.626387v1).

PR #452, opened 2025-02-20
feat: Add instance point converter for copick input

uermel:
Relates to https://github.com/chanzuckerberg/cryoet-data-portal-backend/pull/438

## Description
In order to properly import data from the above linked submission, we need to import instance points from copick files. This PR adds this ability: reuses old `from_copick` converter, but changes the output type to InstancePoint. Each point is considered one instance, and IDs are assigned incrementally. Also adds a test case and test file.


manasaV3:
out of curiousity, If each point is its own instance ie there are no groups, what is the benefit of having this as an instance segmentation over a normal point?

uermel:
Having an explicit ID assigned to the point rather than an implicit one from the list index. The output from their model is actually an instance segmentation mask (i.e. a dense mask with each vesicle's segmentation having a different integer value). Short term, this is easier to support for us. Longer term, we can think about adding instance segmentation masks.

manasaV3 (approved):

PR #455, opened 2025-02-26
fix: Update Ingestion Config 

rainandbare:
Relates to
https://github.com/chanzuckerberg/cryoet-data-portal-backend/pull/453 - 

## Description
Including deposition tags and author kaggle ids


schema/core/v2.0.0/metadata.yaml
```diff
@@ -257,6 +260,9 @@ classes:
       deposition_types:
         exact_mappings:
           - cdp-common:deposition_types
+      deposition_tag:
```
rainandbare, line 263: I am not sure if this should be deposition_tag or just tag as it has been called in the main apiv2 changes

kaloster (approved):

PR #461, opened 2025-03-13
feat: add PerSectionParameters table to apiv2

Bento007:
Relates to https://github.com/chanzuckerberg/cryoet-data-portal/issues/1523

## Description
- Update the apiv2 schema to include PerSectionParameters.
- Update the db_import to generate PerSectionParameters from Tiltseries metadata.
- Regenerated apiv2, and update the schema diagram.


apiv2/schema/schema.yaml
```diff
@@ -1771,6 +1780,111 @@ classes:
...
+        value: PerSectionParameters
+    description: Map individual Frames to a Tiltseries
+    from_schema: cdp-dataset-config
+    mixins:
+      - IDMixin
+    attributes:
+      astigmatic_angle:
+        name: astigmatic_angle
+        description: Angle (in degrees) from reciprocal space X axis to the major axis of defocus.
+        from_schema: cdp-dataset-config
```
manasaV3, line 1797: I think we don't need the `from_schema` here and the other entries below because this is not referencing a definition from else where, but rather defining the attribute here.

manasaV3 (changes requested):

apiv2/database/migrations/versions/20250219_050850_autogenerated.py
```diff
@@ -26,7 +26,8 @@ def upgrade() -> None:
 def downgrade() -> None:
     # ### commands auto generated by Alembic - please adjust! ###
     op.add_column(
-        "frame", sa.Column("raw_angle", sa.DOUBLE_PRECISION(precision=53), autoincrement=False, nullable=True),
+        "frame",
+        sa.Column("raw_angle", sa.DOUBLE_PRECISION(precision=53), autoincrement=False, nullable=True),
```
manasaV3, line 30: Parking the comment here as it is related to migrations. It looks like we are missing the migrations related to table creation for the `PerSectionParameter`.

apiv2/db_import/importer.py
```diff
@@ -254,6 +258,13 @@ def load_func(
                 if tiltseries:
                     tiltseries_obj = tiltseries.import_to_db()
                     tiltseries_cleaner.mark_as_active(tiltseries_obj)
+                    parents = {"run": run_obj, "tiltseries": tiltseries_obj}
```
manasaV3, line 261: nit: Do we need the run object here or could that be referenced from the tiltseries object?

apiv2/db_import/tests/populate_db.py
```diff
@@ -396,6 +399,28 @@ def populate_tiltseries(session: sa.orm.Session) -> Tiltseries:
...
     )
+    frame = Frame(
+        run_id=RUN1_ID,
+        id=FRAME_ID,
+        acquisition_order=0,
+        deposition_id=DEPOSITION_ID1,
+        s3_frame_path="s3://test-public-bucket/30001/RUN1/Frames/frame1",
+        https_frame_path="https://foo.com/30001/RUN1/Frames/frame1",
+    )
+    session.add(frame)
```
manasaV3, line 410: nit: Could this be replaced with the `populate_existing_frames` method below?

apiv2/database/migrations/versions/20250219_050850_autogenerated.py
Bento007: Added it in. I forgot to stage it.

apiv2/schema/schema.yaml
```diff
@@ -1154,7 +1154,7 @@ classes:
     slot_usage:
       primary_author_status:
         description: Indicates whether an author is the main person associated with the corresponding dataset
-      name::
+      name:
```
Bento007, line 1157: @manasaV3 was this a typo?

apiv2/schema/README.md
```diff
@@ -0,0 +1,354 @@
...
+    string kaggle_id
+    string name
+    string email
+    string affiliation_name
+    string affiliation_address
+    string affiliation_identifier
+    boolean corresponding_author_status
+    boolean primary_author_status
+}
+PerSectionParameters {
```
Bento007, line 60: @manasaV3 it's generated now

apiv2/db_import/tests/populate_db.py
Bento007: `populate_existing_frames` does more than what we need so I moved that function back to `test_frame_import.py`

apiv2/db_import/importers/tiltseries.py
```diff
@@ -1,14 +1,65 @@
...
+        self.model_args["tiltseries_id"] = self.input_data["tiltseries"].id
+        run_id = self.input_data["tiltseries"].run_id
+        self.model_args["run_id"] = run_id
+        # query the database for the frame_id using run_id and acquisition_order
+        self.model_args["frame_id"] = (
+            self.config.get_db_session()
+            .query(models.Frame)
+            .filter_by(run_id=run_id, acquisition_order=self.input_data["frame_acquisition_order"])
+            .one()
+            .id
```
manasaV3, line 38: This is neat!

apiv2/schema/schema.yaml
manasaV3: I would guess yes!

apiv2/db_import/tests/populate_db.py
```diff
@@ -396,6 +399,29 @@ def populate_tiltseries(session: sa.orm.Session) -> Tiltseries:
...
+    per_section_parameters = PerSectionParameters(
```
manasaV3, line 411: Is there additional changes to the tiltseries test, to validate the updates to this record? 

As the PerSectionParameter has been configured for cleaning up stale records, should we add a stale record to test that as well?

apiv2/schema/schema.yaml
```diff
@@ -1771,6 +1780,100 @@ classes:
...
+      major_defocus:
+        name: major_defocus
+        description: Defocus (major axis) estimated for this tilt image in Angstrom (underfocus has positive sign).
+        exact_mappings:
+          - per_section_major_defocus
+        range: float
+        required: false
+        unit:
+          symbol: Å
+          descriptive_name: Angstrom
```
manasaV3, line 1815: Just to confirm the units for these fields have been source from Utz?

apiv2/schema/schema.yaml
```diff
@@ -1771,6 +1780,100 @@ classes:
...
+      plural:
+        tag: plural
+        value: PerSectionParameters
+    description: Map individual Frames to a Tiltseries
+    mixins:
+      - IDMixin
+    attributes:
+      astigmatic_angle:
+        name: astigmatic_angle
+        description: Angle (in degrees) from reciprocal space X axis to the major axis of defocus.
```
manasaV3, line 1795: Should this have units?

apiv2/schema/schema.yaml
```diff
@@ -1771,6 +1780,100 @@ classes:
         unit:
           symbol: °
           descriptive_name: degrees
+  PerSectionParameters:
```
Bento007, line 1783: @uermel can you take a look at the units and descriptions for these fields and let me know if they make sense.

apiv2/db_import/tests/populate_db.py
```diff
@@ -438,6 +464,24 @@ def populate_stale_tiltseries(session: sa.orm.Session) -> None:
     )
 
 
+@write_data
+def populate_stale_per_section_parameters(session: sa.orm.Session) -> PerSectionParameters:
+    session.add(
+        PerSectionParameters(
```
manasaV3, line 470: Could we also add a testing for the case where we are deleting a record that was added in the past with incorrect data for a valid run, and the updated data no longer contains that record?

apiv2/db_import/tests/populate_db.py
```diff
@@ -438,6 +464,24 @@ def populate_stale_tiltseries(session: sa.orm.Session) -> None:
...
+            astigmatic_angle=0.1,
+            frame_id=FRAME_ID,
+            major_defocus=0.5,
+            minor_defocus=0.5,
+            phase_shift=0.5,
+            max_resolution=0.5,
+            raw_angle=0.5,
+            tiltseries_id=TILTSERIES_ID,
+            z_index=0,
+        ),
```
manasaV3, line 481: With this addition, we can also test the invalid sibling of a valid run and tiltseries. 
```suggestion
      ),
      PerSectionParameters(
        run_id=RUN1_ID,
        astigmatic_angle=0.1,
        frame_id=1234,
        major_defocus=0.5,
        minor_defocus=0.5,
        phase_shift=0.5,
        max_resolution=0.5,
        raw_angle=0.5,
        tiltseries_id=TILTSERIES_ID,
        z_index=10,
      ),
```

manasaV3 (approved):

PR #464, opened 2025-03-17
fix: Handling default frame sources in tests

manasaV3:
Relates to

## Description
Handles cases where the frame source is default. Treats the list of frame files as an empty list.


uermel (approved):
Great, works now with these changes!

PR #467, opened 2025-03-20
fix: fix errors on apiv2 schema updates

jgadling:
This pr does a couple things:

1. Updates the migration codegen so that it doesn't produce migration files when nothing's changed
2. Updates the linkml schema generation command to run *inside* a docker container, to prevent errors
3. Update the docs to align better with our current dev process.

### DIFF
--- a/apiv2/Makefile
+++ b/apiv2/Makefile
@@ -38,7 +38,7 @@ build: ## Build docker containers
 update-schema: codegen alembic-autogenerate alembic-upgrade-head
 	# $(docker_compose_run) $(CONTAINER) ruff check --fix .
 	$(docker_compose_run) $(APP_CONTAINER) sh -c 'strawberry export-schema main:schema > /app/graphql_api/schema.graphql'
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -61,3 +61,8 @@ push-ingestor-build-apiv2:
 .PHONY: db-import-dataset
 db-import-dataset:
 	AWS_PROFILE=cryoet-dev aws-oidc exec -- docker run --rm=true -v $$PWD/apiv2:/app -e AWS_REGION -e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY -e AWS_SESSION_TOKEN -e AWS_DEFAULT_OUTPUT --network cryoback_default -ti graphql-api python3 -m db_import.importer load --s3-prefix 10001 --import-everything --postgres_url postgresql+psycopg://postgres:postgres@db:5432/cryoetv2 cryoet-data-portal-public http://localhost:8000  --import-depositions --deposition-id 10301 --deposition-id 10303
+
+.PHONY: update-api-schema
+update-api-schema:
+	$(MAKE) -C apiv2 update-schema
+	pre-commit run --all-files
```
manasaV3, line 68: This might not be needed for `--all-files` when only files inside apiv2 would have changed. 

Also, should this be a part of the make command for apiv2?

jgadling: wait actually it *is* part of that make target. It looks like `pre-commit` behaves differently when run from inside vs outside the apiv2 directory

Bento007 (approved):

PR #469, opened 2025-03-21
chore: update cryoet data portal neuroglancer version from 1.0.1 to 1.2.3

seankmartin:
The main goal is to update the [cryoet-data-portal-neuroglancer](https://github.com/chanzuckerberg/cryoet-data-portal-neuroglancer) version to 1.2.1. Overall this should mean that the following are available in a neuroglancer link from the portal after ingestion:
1. Hide the opaque, single-color background of the orthoplanes in the 3D view of the tomogram
2. opacity slider - additive blending
3. Hide shader code from users
4. Disable "pick" for segmentation layer
5. Color legend relating an annotation displayed in 2D or 3D to its respective tab or sidebar entry in Neuroglancer (this is setup to be ready if the change is made to main neuroglancer - won't immediately appear until then)

See the [changelog](https://github.com/chanzuckerberg/cryoet-data-portal-neuroglancer/blob/main/CHANGELOG.md#121-2025-03-21) for what this should add in more detail.

The lock file has also been updated for this.

### DIFF
--- a/ingestion_tools/poetry.lock
+++ b/ingestion_tools/poetry.lock
@@ -331,13 +331,13 @@ crt = ["botocore[crt] (>=1.21.0,<2.0a0)"]
 
 [[package]]
 name = "boto3-stubs"
-version = "1.37.17"
-description = "Type annotations for boto3 1.37.17 generated with mypy-boto3-builder 8.10.1"
+version = "1.37.20"
+description = "Type annotations for boto3 1.37.20 generated with mypy-boto3-builder 8.10.1"
 optional = false
 python-versions = ">=3.8"
 files = [
-    {file = "boto3_stubs-1.37.17-py3-none-any.whl", hash = "sha256:fc28cd71dff999ffb2b39679e1ea976bc4785a388e5a800f0307c171f72ab9e9"},
-    {file = "boto3_stubs-1.37.17.tar.gz", hash = "sha256:92385752b4e1c2c75f7ca7d95b596beddb806e1f03a556a54167361ac9b43de3"},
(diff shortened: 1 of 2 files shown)


uermel:
Hi @seankmartin, I tried generating pre-compute with this update, but ran into an error like this: 
```
cloudvolume.exceptions.SpatialIndexGapError: mesh/10352.64-12940.80_10352.64-12509.44_6901.76-10352.64.spatial was not found.
```

When I tried looking that up I stumbled on [this issue](https://github.com/seung-lab/igneous/issues/177) in igneous by @aranega. Is there something we need to change about precompute generation before applying this PR?

seankmartin:
Hey @uermel , thanks for spotting this!

There were two issues which we've fixed, hopefully that will resolve the problem. The first was:
https://github.com/chanzuckerberg/cryoet-data-portal-neuroglancer/issues/43 -- so the install wasn't pulling the right version

The second was that the patch on top of igneous (from the issue you spotted) that we placed in https://github.com/chanzuckerberg/cryoet-data-portal-neuroglancer/blob/main/cryoet_data_portal_neuroglancer/igneous_patch.py was not getting applied because the version of igneous in the project wasn't correct for the patch to apply to.
We've updated that in https://github.com/chanzuckerberg/cryoet-data-portal-neuroglancer/pull/48

These two combined should hopefully resolve that issue with the meshes, very happy to take another look if not, and thanks again

uermel:
Very cool, thanks @seankmartin. This seems to work now!

uermel (approved):
Confirmed working for precompute generation locally and as part of the AWS Workflow. 

There seems to be a slight bug as described [here](https://github.com/chanzuckerberg/cryoet-data-portal-neuroglancer/issues/49), but it shouldn't be a blocker for this as there are other important updates included here.

PR #471, opened 2025-03-24
feat: Add support for contrast limit

aranega:
## Description

This PR links the contrast limit computation provided by [cryoet-data-portal-neuroglancer](https://github.com/chanzuckerberg/cryoet-data-portal-neuroglancer) `v1.2.1`.  The contrast limit is computed on the MRC file produced during the ingestion, and uses the `gmm` computation method by default.


ingestion_tools/scripts/importers/base_importer.py
```diff
@@ -253,6 +262,10 @@ def get_output_volume_info(self) -> VolumeInfo:
         output_prefix = self.get_output_path()
         return get_volume_info(self.config.fs, f"{output_prefix}.mrc")
 
+    def get_contrast_limits(self, method: Literal["gmm", "cdf"] = "gmm") -> tuple[float, float]:
+        output_prefix = self.get_output_path()
+        return get_volume_contrast_limits(self.config.fs, f"{output_prefix}.mrc", method=method)
```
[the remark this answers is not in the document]
aranega: If that's the case, then I think we perhaps have a problem, when reaching this point, at least in the tests, it seems that the `.zarr` file isn't produced. I'm not sure if that's comes from the mocks in the tests, or if that's a normal behavior (to not have the `.zarr` file produce).

ingestion_tools/scripts/importers/visualization_config.py
```diff
@@ -174,7 +178,11 @@ def _create_config(self, alignment_metadata_path: str) -> dict[str, Any]:
         volume_info = tomogram.get_output_volume_info()
         voxel_size = round(volume_info.voxel_size, 3)
         resolution = (voxel_size * 1e-10,) * 3
-        layers = [self._to_tomogram_layer(tomogram, volume_info, resolution)]
+        t = time()
+        print("Start contrast limit computation for", tomogram)
+        contrast_limits = tomogram.get_contrast_limits()
+        print("Took", (time() - t), "s")
```
seankmartin, line 184: I think we could combine the print for the computed contrast limit and the time taken into one to reduce the number of prints

ingestion_tools/scripts/importers/visualization_config.py
```diff
@@ -174,7 +177,11 @@ def _create_config(self, alignment_metadata_path: str) -> dict[str, Any]:
         volume_info = tomogram.get_output_volume_info()
         voxel_size = round(volume_info.voxel_size, 3)
         resolution = (voxel_size * 1e-10,) * 3
-        layers = [self._to_tomogram_layer(tomogram, volume_info, resolution)]
+        t = time()
+        print("Start contrast limit computation for", tomogram)
```
Bento007, line 181: Does this need to stay in the code? I assume it was here for performance tuning.

Bento007 (changes requested):
I think you left some debugging code

ingestion_tools/scripts/importers/visualization_config.py
aranega: @Bento007 Thanks for the feedback! Actually, we added some feedbacks as the contrast limit can take a little bit of time depending on the method, or the level of the pyramid that is used as input for the computation. The idea was to not have just the scripts that hangs for few seconds without any information about why it is. If that's something that you feel is not necessary, we can remove it.

Bento007: No, that's fine. It would be worth adding a comment as to why we add this metric, since we aren't doing it anywhere else.

Bento007 (approved):

ingestion_tools/scripts/importers/visualization_config.py
aranega: I updated the PR with the comment :)

Bento007 (approved):

PR #477, opened 2025-04-01
chore: Switch to pulling the platformics library from pypi.

jgadling:
We've been using a forked version of platformics until now. All of the nice improvements we've made to our forked copy of platformics have been merged upstream, so let's start pulling in the upstream library instead of maintaining a fork!


apiv2/pyproject.toml
```diff
@@ -52,6 +50,7 @@ strawberry-graphql = "0.257.0"
 typer = "^0.12.3"
 libcst = "^1.4.0"
 s3fs = "^2024.9.0"
+platformics = "^1.0.2"
```
jgadling, line 53: This is where we pull in the upstream package!

PR #479, opened 2025-04-08
chore: update top level readme.md and remove reference to API V1

Bento007:
## Description
- now that apiV1 is deprecated and unused there is no reason to keep it's schema and build recipes.
- Verified the changes by running [make build](https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/a46df8985fe081147484fdf2310d3efe0aca88c9/schema/Makefile#L57), and verifying that no error or files changes


manasaV3 (approved):
Thanks for cleaning up the readme along with the unused schema! 🚢

PR #498, opened 2025-06-11
Feat/multi tomogram spacing

seankmartin:
## Description

Tomograms can be generated at different voxel spacings for the same run. Previously, these tomograms had no neuroglancer config generated for them. This PR updates that to produce a neuroglancer config for these tomograms with multiple voxel spacing.

### Config generation - different data scales and viewer scales

A change was first required in the dependent project, which has been done at https://github.com/chanzuckerberg/cryoet-data-portal-neuroglancer/pull/62. This would allow generating configurations where the data in the layers could have different data resolutions to the global viewer resolution. This would let a tomogram of say 1nm resolution determine that each unit in the viewer is 1nm in size. But then to also allowing adding two other data layers to the viewer, each at 3nm voxel spacing, so different to the global spacing. This difference in data layer resolution and global dimension resolution was not previously supported in the config generation.

TODO this means the version in this PR which points to cryoet-data-portal-neuroglancer in the pyproject.toml needs to be bumped once PR62 has been merged.

### Config generation - picking up annotations from different voxel spacings

The second main change was in this backend repository. Because the annotations for each tomogram in the same run are the same, there was no reason to run precompute for each different voxel spacing. Instead when the visualization config was generated, it should search across the different available voxel spacing folders in the output precompute path and grab any precomputed annotations.
In this way, the annotations could be precomputed just for `VoxelSpacing10.012`, but then the neuroglancer config for any tomograms in `VoxelSpacing4.990` would see that there are precomputed annotations in `VoxelSpacing10.012` -- and so include them in a neuroglancer vis config like `VoxelSpacing4.990/NeuroglancerPrecompute/TOMO_ID-neuroglancer_config.json`.

### Config generation - linking together the different resolutions of data

One other change is to link the backend to the changes in the dependent `cryoet-data-portal-neuroglancer` repository. This mostly involves keeping track of the voxel spacing that relates to each annotation (inferred by parsing the `VoxelSpacingXXX` folder name). This voxel spacing is then passed to the neuroglancer config generation. In addition, the ratio between the tomogram spacing and the annotations spacing is tracked because this is needed to correctly setup the initial camera parameters in neuroglancer regardless of the voxel spacing of the tomograms.

### Tests

One test has been removed, because now neuroglancer vis configs get generated for tomograms even if the tomogram has `is_visualization_default = False`. Instead the test for the tomograms being generated runs over both True and False for this metadata. 
One test has been updated because the annotation config generation now checks for the existence of files in the output bucket. If the files don't exist, they get skipped. The tests didn't need to handle this previously. So now the test has been updated to produce a mock output folder with an `info` for each annotation which is supposed to be in the output bucket at the neuroglancer precompute path.

### Refactor - get_annotation_neuroglancer_precompute_path

The `get_annotation_neuroglancer_precompute_path` has been created so that it can be used publicly. Previously the `get_annotation_layer_info` function was repeating some logic of the private `_get_neuroglancer_precompute_path` function. Now the logic has been extracted into a public function that both the `get_annotation_layer_info` and the `_get_neuroglancer_precompute_path` can use


kaloster (approved):

seankmartin:
This has been updated to account for PRs 499 and 500 cc @uermel @aranega

PR #505, opened 2025-06-17
fix: doi publication validation bug

daniel-ji:
When making network requests to api.crossref.org to validation publication DOIs, we currently try to make a `HEAD` request to a URL along the lines of `https://api.crossref.org/works/{doi}/agency`. However, an 405 error (Method Not Allowed) status now appears, likely because there should only be `GET` requests getting sent to this endpoint. As a fix, we just make a `HEAD` request to `https://api.crossref.org/works/{doi}`, which properly validates the DOI and does not give this 405 error. 

For reference: 
https://api.crossref.org/swagger-ui/index.html#:~:text=Headers%20only%3A%20you,org/members/98%22


uermel (approved):
⛴️

PR #520, opened 2025-07-22
chore: Updating the ingestion config  and schema

manasaV3:
Relates to 
- https://czi.atlassian.net/browse/VC-3381
- https://czi.atlassian.net/browse/VC-3380

## Description

Updates the ingestion config schema and the values for the ingestion configs to support values such as a assay, development_stage, disease.


ingestion_tools/dataset_configs/10006.yaml
```diff
@@ -1,245 +1,254 @@
...
+    - ORCID: 0000-0002-4551-1811
+      name: Christopher Neufeldt
+    - name: Berati Cerikan
+    - ORCID: 0000-0002-3972-8854
+      name: John Lu
+    - ORCID: 0000-0001-8560-7407
+      name: Julia Peukes
+    - ORCID: 0000-0002-4632-9122
+      name: Xiaoli Xiong
+    - name: Hans-Georg Kräusslich
```
uermel, line 37: The automatic addition seems to have un-escaped some strings. I'm not 100% sure why they were escaped before, but are we sure that it is fine to un-escape these?

ingestion_tools/dataset_configs/10446.yaml
```diff
@@ -1,610 +1,603 @@
...
+    annotation_method: hybrid curation using CellCanvas, copick, slabpick, deepfindET, topaz, minislab curation and cryoSPARC
+    annotation_object:
+      state: tetrameric
+      id: UniProtKB:P10537
+      name: Beta-amylase
+    annotation_publications: 10.1101/2024.11.04.621686, 10.1101/2024.11.04.621608, EMD-30405
+    annotation_software: CellCanvas + ChimeraX-copick 0.5.0 + copick-live + pytom-match-pick 0.6.0 + slabpick 0.1.0 + deepfindET
+      0.4.0 + topaz 0.2.5 + cryoSPARC 4.4.1
+    authors: *id001
+    dates: *id002
```
uermel, line 233: In some cases YAML references seem to be respected (although it is not ideal they are renamed to useless names), but in this example the `annotation_method` was for example copied into every annotation rather than using the reference. 

This is not great, because now manual changes will be necessary in a lot of places if anything needs to be updated.

uermel (approved):
Looks fine in general, my main concern are the changes to the manual configs, because they dereference some YAML-refs, rename all YAML references and un-escape some escaped text. 

Changes to the YAML-references are not ideal if I ever have to look at these files again as a human. Is there any way of not breaking all of them? 

I will approve here to get this out, but with a bit of hesitation.

ingestion_tools/curation.tsv
Bento007, line 1: Can you add more information on the origin of this file for future editing, or if you could update one of the JIRA issue or PR description with more context.

Bento007 (changes requested):

ingestion_tools/curation.tsv
uermel: It's manual curation I did on all datasets.

uermel (approved):
👍

ingestion_tools/curation.tsv
Bento007: it's a one off that shouldn't need modifying in the future?

uermel: This is for fullfilling MDR requirements. I don't know how MDR requirements will change in the future. Might need modification at some stage.

PR #521, opened 2025-07-23
chore: Updating dataset metadata ingestion to handle cross modality data

manasaV3:
Relates to

### DIFF
--- a/ingestion_tools/scripts/importers/dataset.py
+++ b/ingestion_tools/scripts/importers/dataset.py
@@ -60,11 +60,14 @@ class DatasetImporter(BaseImporter):
             if key not in data:
                 data[key] = {}
             if not data[key].get("name"):
-                data[key]["name"] = "not_reported"
+                data[key]["name"] = "unknown" if key == "development_stage" else "not_reported"
             if key == "organism":
                 if not data[key].get("taxonomy_id"):
                     data[key]["taxonomy_id"] = None
             elif not data[key].get("id"):
-                data[key]["id"] = "not_reported"
+                if key == "development_stage":
+                    data[key]["id"] = "unknown"
+                else:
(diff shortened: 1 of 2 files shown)


ingestion_tools/scripts/importers/dataset.py
```diff
@@ -36,3 +40,31 @@ def load_extra_metadata(self) -> dict[str, dict[str, str]]:
...
+        for key in keys:
+            if key not in data:
+                data[key] = {}
+            if not data[key].get("name"):
+                data[key]["name"] = "not_reported"
+            if key == "organism":
+                if not data[key].get("taxonomy_id"):
+                    data[key]["taxonomy_id"] = "not_reported"
+            elif not data[key].get("id"):
+                data[key]["id"] = "not_reported"
```
uermel, line 68: For `development_stage` this should set the value to `unknown` to conform to XMS-1.1.0

uermel (changes requested):

ingestion_tools/scripts/importers/dataset.py
uermel: Also, on second though, is `taxonomy_id` not an `int` anymore? In the schema PR it seems to by, so this may cause issues here.

manasaV3: > For development_stage this should set the value to unknown to conform to XMS-1.1.0

For `developmental_stage` the values in config already has `unkown` set as an acceptable fallback. 

ex: https://github.com/chanzuckerberg/cryoet-data-portal-backend/pull/520/files#diff-179989eb2c11a44b5e11ad4c986b30c4ebc812348d7693a57c11adceeb48d802R156

uermel (approved):
LGTM!

PR #522, opened 2025-07-23
chore: Updating dataset db ingestion for cross modality schema

manasaV3:
Relates to


apiv2/schema/schema.yaml
```diff
@@ -1259,6 +1259,30 @@ classes:
...
+        description: Ontology identifier for the development stage
+        range: string
+      disease_name:
+        name: disease_name
+        description: Name of the disease associated with the biological sample used in a CryoET study, e.g. Alzheimer’s disease
+        range: string
+      disease_ontology_id:
+        name: disease_ontology_id
+        description: Ontology identifier for the disease associated with the biological sample used in a CryoET study
+        range: string
```
uermel, line 1285: Any reason why these new fields are not required in the DB? They are required in XMS-1.1.0. 

Also, the older fields here seem to import from another schema using `from_schema: cdp-dataset-config`. Would it make sense to maintain that pattern?

uermel (approved):
Looks good to me. I saw TODOs to make the old fields required, and the new fields don't  seem to be required either. 

Is there a reason for not doing that at the moment?

apiv2/schema/schema.yaml
manasaV3: The requirement of the field not being null is currently managed programatically. 

As we are adding this column to an existing table, we can't mark it as required aka non null, as at the time of addition of the columns they will not have values. The way to implement that at the db level is, 
1. Add a nullable column to the existing table
2. Ensure all the columns have non-null values set
3. Add the non-null constraint on the column.

Bento007 (approved):

PR #542, opened 2025-11-04
feat: cryolens dataset

uermel:
Ingestion config for cryoLens dataset.


ingestion_tools/dataset_configs/10454.yaml
```diff
@@ -0,0 +1,3050 @@
...
+    authors: *id001
+    ctf_corrected: true
+    dates: *id002
+    fiducial_alignment_status: NON_FIDUCIAL
+    is_visualization_default: true
+    offset:
+      x: 0
+      y: 0
+      z: 0
+    processing: raw
```
daniel-ji, line 3037: should this be denoised?

ingestion_tools/dataset_configs/10454.yaml
```diff
@@ -0,0 +1,3050 @@
...
+        PDB-6KLH, PDB-6KRK, PDB-6KSP, PDB-6LMT, PDB-6LX3, PDB-6LXK, PDB-6LXV, PDB-6M04,
+        PDB-6MRC, PDB-6PIF, PDB-6TA5, PDB-6TAV, PDB-6TPS, PDB-6U8Q, PDB-6UP6, PDB-6VGR,
+        PDB-6VN1, PDB-6VZ8, PDB-6W6M, PDB-6WZT, PDB-6X5Z, PDB-6XF8, PDB-6YT5, PDB-6Z3A,
+        PDB-6Z6O, PDB-6Z80, PDB-6ZIU, PDB-6ZQJ, PDB-7AMV, PDB-7B5S, PDB-7B7U, PDB-7BKC,
+        PDB-7BLQ, PDB-7BLR, PDB-7DD9, PDB-7E1Y, PDB-7E6G, PDB-7E8H, PDB-7EEP, PDB-7EGD,
+        PDB-7EGE, PDB-7EGQ, PDB-7ETM, PDB-7EY7, PDB-7JSN, PDB-7K5X, PDB-7KDV, PDB-7KFE,
+        PDB-7KJ2, PDB-7MEI, PDB-7NHS, PDB-7NIU, PDB-7NYZ, PDB-7O01, PDB-7Q21, PDB-7QJ0,
+        PDB-7R04, PDB-7S7K, PDB-7SFW, PDB-7SGM, PDB-7SHK, PDB-7SN7, PDB-7T3U, PDB-7VTQ,
+        PDB-7WBT, PDB-7WOO
+    dataset_description: "The CryoLens training dataset consists of 14,366 synthetic\
```
daniel-ji, line 2945: there are 14,376 runs it seems? i.e. running `aws s3 ls s3://cryoetportal-rawdatasets-dev/CZI/20250922_cryolens_training_set/ExperimentRuns/ | grep -c 'tomo-0000-very-crowded-'` gives you that 14,376 now, and it seems like there's no `exclude` blocks here. but also, on prod, it says there's only 200 runs.

daniel-ji (commented):
Regarding the annotations--should we switch their annotation_object ids to all be PDB? Or is there some that we should keep as is? If we switch all to PDB, we might have to also change the annotation_object name for ~20 annotations that are currently GO annotation_objects. I haven't really reviewed the annotation objects, wanted to wait on your call to see what we should do with their ids/names. Thanks!

Each run has a multilabel ground-truth segmentation (Segmentations/...multilabel.zarr) that isn't ingested (points only). The polnet datasets 10439/10441 ingest theirs as SemanticSegmentationMask. Should it be ingested here, or is points-only intended?

PR #649, opened 2026-04-10
fix: improved ingestion config validation & updated template.yaml

daniel-ji:
- check the type of the dataset and deposition source literal value is correct
- only run the workflow on the changed dataset config file if that was the only thing changed with regards to ingestion config validation 
- update template.yaml


schema/ingestion_config/v1.0.0/ingestion_config_models_extended.py
```diff
@@ -719,6 +719,19 @@ def validate_sources(source_list: List[StandardSource] | List[VoxelSpacingSource
         raise ValueError(total_errors)
 
 
+def validate_literal_values(source_list: List[StandardSource], expected_type: type, entity_name: str) -> None:
```
PALLAVIKHEDLE, line 722: This raises on the first bad value and stops - but `validate_sources` above collects all errors into a list and raises once at the end so a config with multiple bad literal values reports all of them at once rather than one per CI run.

PALLAVIKHEDLE (approved):
Nice work, good optimization. added comments - Could you also add unit tests for validate_literal_values?

LGTM!

schema/ingestion_config/v1.0.0/ingestion_config_models_extended.py
daniel-ji: fixed, thanks!

.github/workflows/ingestion-config-validate.yaml
```diff
@@ -41,9 +43,44 @@ jobs:
...
+        run: |
+          # For workflow_dispatch, validate everything
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            echo "validate_all=true" >> $GITHUB_OUTPUT
+            exit 0
+          fi
+
+          # Determine base commit for diff
+          if [ "${{ github.event_name }}" = "pull_request" ]; then
+            BASE=${{ github.event.pull_request.base.sha }}
```
PALLAVIKHEDLE, line 57: Nit: GitHub recommends using env: variables instead of inlining ${{ }} expressions directly in run: scripts - GitHub substitutes these as raw text before the shell runs, which can be a script injection risk.
Ref: [GitHub security hardening docs](https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable)

daniel-ji: Sorry, didn't see this - fixed now in #667. Thank you!

PALLAVIKHEDLE: Awesome, Thanks!

PR #666, opened 2026-04-23
fix: 10440.yaml and 10453.yaml files, bug in enqueuing runs

daniel-ji:
Reingested 10453 annotations due to wrong binning level in the config.

When enqueuing runs, flags after a filter should not be deleted, for example, the `--import*` flags will be ignored in this case because it comes after a `filter` flag:

`python enqueue_runs.py queue ../dataset_configs/10453.yaml cryoetportal-rawdatasets-dev cryoet-data-portal-staging --ecr-tag 10453-fix-1 --filter-run-name "^tomogram_008$" --import-annotations --import-annotation-metadata --import-annotation-viz`

### DIFF
--- a/ingestion_tools/dataset_configs/10440.yaml
+++ b/ingestion_tools/dataset_configs/10440.yaml
@@ -422,7 +422,6 @@ depositions:
     deposition_title: CZII - CryoET Object Identification Challenge
     deposition_types:
     - dataset
-    - annotation
   sources:
   - literal:
       value:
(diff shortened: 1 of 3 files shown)


PALLAVIKHEDLE (commented):
Good catch - Can you confirm the annotations in this config don't belong to deposition 10310? The config still has annotation sources so just want to make sure removing annotation from 10310's deposition_types is intentional and they belong to a different deposition.

uermel (approved):

PR #668, opened 2026-04-25
fix: don't rescale annotation if dims are the same

daniel-ji:
Currently, if you set `rescale: true` but the annotation dimensions and the tomogram dimensions are the same, an unnecessary rescale will be done. This wastes CPU and memory loading the full volume into Python and applying a no-op resize calculation.

In fixing this, we discovered another bug that was using the n{x,y,z}start value from MRC headers for volume dimension calculation, which is incorrect (should only be using the nx/ny/nz value). Resolved here as well. Added a test to ensure this fix works.

### DIFF
--- a/ingestion_tools/scripts/importers/annotation.py
+++ b/ingestion_tools/scripts/importers/annotation.py
@@ -14,7 +14,7 @@ from common.config import DepositionImportConfig
 from common.finders import BaseFinder, DepositionObjectImporterFactory, SourceGlobFinder, SourceMultiGlobFinder
 from common.fs import FileSystemApi
 from common.id_helper import IdentifierHelper
-from common.image import check_mask_for_label, make_pyramids
+from common.image import check_mask_for_label, get_volume_info, make_pyramids
 from common.metadata import AnnotationMetadata
 from importers.alignment import AlignmentImporter
 from importers.base_importer import BaseImporter
@@ -271,6 +271,16 @@ class VolumeAnnotationSource(BaseAnnotationSource):
         )
         return dims
 
+    def _get_resize_target(self) -> tuple[int, int, int] | None:
+        """Return target dims when rescaling is needed; None if rescale is off
+        or the source already matches the target dims (no-op resize)."""
+        if not getattr(self, "rescale", False):
+            return None
+        target = self.get_output_dim()
+        info = get_volume_info(self.config.fs, self.path)
+        src = (info.zend, info.yend, info.xend)
+        return None if src == target else target
+
     def get_metadata(self, output_prefix: str) -> list[dict[str, Any]]:
         metadata = [
             {
@@ -310,7 +320,7 @@ class SegmentationMaskAnnotation(VolumeAnnotationSource):
             raise ValueError("Thresholding and selecting by label are mutually exclusive")
 
     def convert(self, output_prefix: str):
-        output_dims = self.get_output_dim() if self.rescale else None
+        output_dims = self._get_resize_target()
         return make_pyramids(
             self.config.fs,
             self.get_output_filename(output_prefix),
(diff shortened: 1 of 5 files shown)


ingestion_tools/scripts/importers/annotation.py
```diff
@@ -277,8 +277,8 @@ def _get_resize_target(self) -> tuple[int, int, int] | None:
         if not getattr(self, "rescale", False):
             return None
         target = self.get_output_dim()
-        info = get_volume_info(self.config.fs, self.path)
-        src = (info.zend, info.yend, info.xend)
+        d = get_volume_info(self.config.fs, self.path).get_dimensions()
+        src = (d["z"], d["y"], d["x"])
```
uermel, line 281: Can you explain this change? what offset is this referring to?

daniel-ji: sorry, it shouldn't be referring to any offset, it should just be the dimensions of the MRC volume, and it seems like the get_dimensions gives us the volume dims (the previous commit in this PR was not factoring for the potential offset that might occur in the MRC nxstart / nystart / nzstart fields, but this current code is good now)

uermel: But we never previously considered n{x,y,z}start anywhere. Why should we? 

We are interested in comparing the voxel-dimensions of the volume, which are given in nx, ny, and nz fields of the header.

daniel-ji: Yes sorry, I see much of the validation tests and a small bit of the importer (MultiLabelMaskConverter) looks at nx/ny/nz. But to clarify, we do account for n{x,y,z} start:

In `class MRCReader(VolumeReader)`:

https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/35add06d4cea837b623d1b750c781e1090459b27/ingestion_tools/scripts/common/image.py#L284-L295

And when `get_dimensions` is called, it subtracts the provided n{x,y,z}start value from the n{x,y,z} end value:

https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/35add06d4cea837b623d1b750c781e1090459b27/ingestion_tools/scripts/common/image.py#L41-L60

And this is currently used as part of the neuroglancer config generation process in `VisualizationConfigImporter` `_to_tomogram_layer`:

https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/35add06d4cea837b623d1b750c781e1090459b27/ingestion_tools/scripts/importers/visualization_config.py#L87-L108

And when storing tomogram metadata as part of Alignments in `AlignmentImporter`:

https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/35add06d4cea837b623d1b750c781e1090459b27/ingestion_tools/scripts/importers/alignment.py#L118-L125

Is this as intended? Thanks!

uermel: Ah, I never realized this. It's a misinterpretation of what those header fields actually mean. I guess luckily they've always been zero. These fields are basically the coordinate origins, they shouldn't have any bearing on the actual volume size.

uermel: compare https://www.ccpem.ac.uk/mrc-format/mrc2014/

daniel-ji: I see, ok so I can fix this by just dropping 

```
     # start coords 
     xstart: int 
     ystart: int 
     zstart: int 
  
     # end coords 
     xend: int 
     yend: int 
     zend: int 
```

and swapping it for just `xdim`, `ydim`, `zdim` class variables, representing the size? And update the instantiation calls and get_dimensions function accordingly. Thoughts?

this is also used by the Zarr volumes, but it's hard coded for the start coords to be all 0, so not much of an issue there.

uermel: yes, that should be good. I just don't know where xend is even coming from?

daniel-ji: The instantiation here passes nx as xend

https://github.com/chanzuckerberg/cryoet-data-portal-backend/blob/35add06d4cea837b623d1b750c781e1090459b27/ingestion_tools/scripts/common/image.py#L284-L295



and because there's a`@dataclass` decorator on the VolumeInfo class, the init function is written automatically so it corresponds to the `xend` field in the VolumeInfo class

uermel (approved):

PR #674, opened 2026-05-01
fix: fix identified object re-ingestion crash and run filtering

PALLAVIKHEDLE:
- **Add `_load_ids_for_container` override** on`IdentifiedObjectIdentifierHelper` to handle the flat directory structure. Instead of parsing the directory name, it reads the integer `identifier` directly from the stored metadata JSON. Falls back to `100` with a `try/except` for backward compatibility with any existing metadata files that still contain the old file-path string.
- **Fix CSV column filter** in `import_item`: `run_name` → `run_identifier`. Now correctly filters rows to the current run and drops the `run_identifier` column from the output JSON.
- **Fix stored identifier** in `import_metadata`: `self.path` → `self.identifier`. The metadata file now stores the actual integer ID, which the new `_load_ids_for_container` reads back on re-ingestion.

### DIFF
--- a/ingestion_tools/scripts/importers/identified_object.py
+++ b/ingestion_tools/scripts/importers/identified_object.py
@@ -35,7 +35,8 @@ class IdentifiedObjectIdentifierHelper(IdentifierHelper):
             except (ValueError, TypeError):
                 identifier = 100
             cls.cached_identifiers[current_ids_key] = identifier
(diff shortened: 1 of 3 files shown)


uermel (changes requested):
I tested this with 10479, this caused all identified objects to be present for all runs, i.e. selection is somehow broken now.

uermel (approved):

PR #720, opened 2026-06-25
fix: prevent s3 download race conditions (breaks source validation)

daniel-ji:
To be merged in before #676.

During S3 source validation, sometimes an *.mdoc file will end up half-downloaded due to race conditions from multiple pytest-xdist workers that are all trying to download the same file from S3 into the same temp path. A worker may then read a half-written / empty file, causing flaky validation tests. Fix the root problem here by downloading the entire file to a random temp file, and then atomically copying it over so there's never a half-written downloaded file.

### DIFF
--- a/ingestion_tools/scripts/common/fs.py
+++ b/ingestion_tools/scripts/common/fs.py
@@ -7,6 +7,7 @@ from abc import ABC, abstractmethod
 from hashlib import md5
 from io import TextIOBase
 from typing import Any
+from uuid import uuid4
 
 import boto3
 from s3fs import S3FileSystem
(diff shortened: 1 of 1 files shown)


uermel (approved):

ingestion_tools/scripts/common/fs.py
```diff
@@ -106,7 +107,11 @@ def localreadable(self, path: str) -> str:
             if remote_checksum == local_checksum:
                 print(f"Skipping re-download of {path}")
                 return local_dest_file
-        self.s3fs.get(path, local_dest_file)
+        # Download to a unique temp file then atomically move it into place to prevent race conditions with half-written files.
+        os.makedirs(os.path.dirname(local_dest_file), exist_ok=True)
+        tmp_dest_file = f"{local_dest_file}.{os.getpid()}.{uuid4().hex}.part"
+        self.s3fs.get(path, tmp_dest_file)
+        os.replace(tmp_dest_file, local_dest_file)
```
uermel, line 114: Does it really address the race? Couldn't it still race during the move?

daniel-ji: I believe the replace is atomic, done in one go - from not existing to full file contents, no in between.

https://docs.python.org/3/library/os.html#:~:text=os.replace,a%20POSIX%20requirement).

uermel: Makes sense, good to go then!

PR #726, opened 2026-06-27
fix: show errors on submitted jobs and fail fast on missing aws infra

daniel-ji:
When submitting jobs with `enqueue_runs.py`, make sure that any failed job submission shows an error (so that when a job doesn't even get submitted properly to AWS Step Functions, you get an error, and aren't left wondering why your job is not showing up on the Executions list of jobs).

Also show an error when AWS infra doesn't exist when you try to submit a job (an even earlier check to surface any errors).

### DIFF
--- a/ingestion_tools/scripts/enqueue_runs.py
+++ b/ingestion_tools/scripts/enqueue_runs.py
@@ -3,7 +3,7 @@ import logging
 import os.path
 import re
 import time
-from concurrent.futures import ProcessPoolExecutor, wait
+from concurrent.futures import ProcessPoolExecutor, as_completed
 from functools import partial
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


ingestion_tools/scripts/enqueue_runs.py
```diff
@@ -86,6 +86,19 @@ def cli(ctx):
...
+    """Wait for all submitted jobs, re-raising any errors instead of swallowing them."""
+    errors = 0
+    for future in as_completed(futures):
+        try:
+            future.result()
+        except Exception:
+            errors += 1
+            logger.exception("Failed to submit job")
+    if errors:
+        raise RuntimeError(f"{errors} of {len(futures)} job submissions failed; see logs above.")
```
uermel, line 99: This means that the CLI will need to continue running for the entirety of ingestion?

daniel-ji: the future.result() does not wait for the full job to finish because the future is the future returned by the process pool submit call. that future is the result of run_job, which just returns details of the job submission, not actual job results (and will also error if the job submission failed).

this pr is for issues like when you try to run db import with the wrong aws profile for the environment, so the job submission fails because that environment's infra (wdl files, state machine) doesn't exist in the account the credentials point to (might be speaking from personal experience lol). https://docs.aws.amazon.com/boto3/latest/reference/services/stepfunctions/client/start_execution.html#:~:text=RETURNS%3A-,Response%20Syntax,-%7B%0A%20%20%20%20%27executionArn%27

uermel: ah, that sounds reasonable.

uermel (approved):

PR #727, opened 2026-06-29
feat: Update unroofing configs.

uermel:
Relates to

## Description
Replace tomograms and downsampled tiltseries by reprocessed variants.

### DIFF
--- a/ingestion_tools/dataset_configs/10427.yaml
+++ b/ingestion_tools/dataset_configs/10427.yaml
@@ -1,17 +1,22 @@
 alignments:
 - metadata:
-    format: IMOD
+    format: ARETOMO3
+    is_portal_standard: true
     method_type: fiducial_based
   sources:
-  - source_multi_glob:
-      list_globs:
-      - alignmentfiles/{run_name}_binalign8_fid.tlt
-      - alignmentfiles/{run_name}_binalign8_fid.xf
+  - source_glob:
+      list_glob: reprocessed/{run_name}_*/{run_name}_*.aln
 annotations: []
 collection_metadata:
 - sources:
   - source_glob:
       list_glob: mdocs/{mdoc_name}
+ctfs:
+- metadata:
+    format: CTFFIND
+  sources:
+  - source_glob:
+      list_glob: reprocessed/{run_name}_*/{run_name}_*_CTF.txt
 dataset_keyphotos:
 - sources:
   - literal:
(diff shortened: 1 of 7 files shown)


ingestion_tools/dataset_configs/10427.yaml
```diff
@@ -292,7 +297,7 @@ tiltseries:
       model: TITAN KRIOS
     microscope_optical_setup:
       energy_filter: Gatan Quantum LS
-    pixel_spacing: 8.627
+    pixel_spacing: 2.157
```
daniel-ji, line 300: this pixel_spacing appears to be stale (from the older IMOD spacings, but the AreTomo3 pixel spacings are slightly different). This value should be `2.165` based on the headers.  Applies for this dataset (10427) and also 10431, 10432, 10433, 10434, 10435

for example see `s3://cryoetportal-rawdatasets-dev/kem_sochacki_06_2024/grid02/reprocessed/tomo_c2c1_11_16886/tomo_c2c1_11_16886.mrc`

ingestion_tools/dataset_configs/10427.yaml
```diff
@@ -325,24 +330,108 @@ tomograms:
...
     is_visualization_default: true
+    offset:
+      x: 0
+      y: 0
+      z: 0
+    processing: filtered
+    reconstruction_method: WBP
+    reconstruction_software: AreTomo3 v2.1.10
+    tomogram_version: 1
+    voxel_spacing: 8.627
```
daniel-ji, line 350: this voxel_spacing appears to be stale (from the older IMOD voxel spacings, but the AreTomo3 voxel spacings are slightly different). This value should be `8.66` based on the headers.  Applies for this dataset (10427), not only this tomogram, but the other two tomogram entries here, and also for 10431, 10432, 10433, 10434, 10435.

example file: `s3://cryoetportal-rawdatasets-dev/kem_sochacki_06_2024/grid02/reprocessed/tomo_c2c1_11_16886/tomo_c2c1_11_16886_ts_Vol.mrc`

daniel-ji (changes requested):

ingestion_tools/dataset_configs/10427.yaml
uermel: no, that's incorrect. They were processed at a uniform spacing even though that's not correct. The values in the configs reflect the actual correct ones.

ingestion_tools/dataset_configs/10427.yaml
uermel: No, same as above.

daniel-ji (approved):

PR #738, opened 2026-07-09
feat: FISE and PACE tomo datasets

uermel:
Relates to

## Description
In vitro ribosome datasets collected using FISE (EMPIAR-10304) and PACE (EMPIAR-10985).

### DIFF
--- a/ingestion_tools/dataset_configs/10498.yaml
+++ b/ingestion_tools/dataset_configs/10498.yaml
@@ -1,0 +1,353 @@
+# CryoET Data Portal Ingestion Configuration — Dataset 10498 / Deposition 10355
+# EMPIAR-10985 (PACE-tomo of 70S ribosomes), CZII reprocessing benchmark.
+# Built from the pipeline stub (cryoet-data-portal-pipeline stage_upload) + pipeline_configs/10498.yaml.
+version: 1.1.0
+
+standardization_config:
+  deposition_id: 10355
+  source_prefix: EMPIAR/EMPIAR-10985/
+  run_data_map_file: run_to_data_map.tsv
+
+datasets:
+- metadata:
+    dataset_identifier: 10498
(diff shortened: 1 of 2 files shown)


ingestion_tools/dataset_configs/10498.yaml
```diff
@@ -0,0 +1,353 @@
...
+    authors: *tomo_authors
+    dates: *dates
+    voxel_spacing: 10.005
+    is_visualization_default: true
+  sources:
+  - source_glob:
+      list_glob: copick/ExperimentRuns/{run_name}/VoxelSpacing*/**/aretomo3-miss-aligned-global-ctf-deconvolved.zarr
+# 3) Warp miss-aligned global (filtered, Fourier-space)
+- metadata:
+    reconstruction_software: WarpTools
```
daniel-ji, line 245: Warp? as per backend convention? (also for processing_software)

ingestion_tools/dataset_configs/10498.yaml
daniel-ji, line 1: should the align_software also include AreTomo3? for both yaml files?

daniel-ji, line 1: Also, do we want the default viz to be on a denoised tomo? like isonet2? or do we wanna keep ctf deconvolved because there's multiple denoised and we don't want to play sides?

daniel-ji (approved):

ingestion_tools/dataset_configs/10498.yaml
uermel: what convention?

ingestion_tools/dataset_configs/10498.yaml
uermel: no.

uermel: yes.

PR #747, opened 2026-07-15
feat: HeLa under GTPP stress. 

uermel:
Relates to

## Description
Deposition config for EMPIAR-12894.

### DIFF
--- a/ingestion_tools/dataset_configs/10511.yaml
+++ b/ingestion_tools/dataset_configs/10511.yaml
@@ -1,0 +1,371 @@
+# CryoET Data Portal Ingestion Configuration
+# Dataset 10511 / Deposition 10360 / EMPIAR-12894
+# Mitochondria under folding stress (in situ cryo-ET) — Jung et al., Sci. Adv. 2025 (10.1126/sciadv.adw6064)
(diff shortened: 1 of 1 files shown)


ingestion_tools/dataset_configs/10511.yaml
```diff
@@ -0,0 +1,371 @@
...
+    cell_component:
+      name: mitochondrion
+      id: GO:0005739
+    assay:
+      name: microscopy assay
+      id: EFO:0002909
+    development_stage:
+      name: unknown
+      id: unknown
+    disease:
```
daniel-ji, line 90: disease is set to 'normal' (PATO:0000461), but both existing HeLa portal configs (10475, 10476 - the sibling FIB-milled HeLa mito-stress studies) use 'cervical carcinoma' (MONDO:0005131). is this intended? thanks!

daniel-ji (approved):
just one question, otherwise looks good!

ingestion_tools/dataset_configs/10511.yaml
uermel: Ah, that's actually a mistake on the other two.

daniel-ji: ah i see, do you want me to pr and re-ingest/re-dbimport the fixed versions of the other datasets then?

